PR #2, opened 2018-10-01
Enable msi creation and signing

soupdiver:


requilence:
@soupdiver looks like SSH key is not working for the Win-builder server: https://circleci.com/gh/cloudradar-monitoring/cagent/25

PR #8, opened 2018-10-20
Windows: Perfcounters & Free memory

requilence:
I have added the perfcounter methods taken from https://github.com/influxdata/telegraf/blob/master/plugins/inputs/win_perf_counters

I rewrite them slightly because telegraf used it in a different manner (fetch the metrics periodically). In our case, we have a global timer, so don't need it.

Currently(in this PR) perfcounter only used to get free memory(`\Memory\Free & Zero Page List Bytes`)

### DIFF
--- a/mem.go
+++ b/mem.go
@@ -33,7 +33,6 @@ func (ca *Cagent) MemResults() (MeasurementsMap, error) {
 		results["buff_B"] = nil
 		results["available_B"] = nil
 	} else {
-
 		results["total_B"] = memStat.Total
 		results["used_B"] = memStat.Used
 		results["free_B"] = memStat.Free
--- a/mem_windows.go
+++ b/mem_windows.go
@@ -10,11 +10,11 @@ import (
 
 	"github.com/shirou/gopsutil/mem"
 	log "github.com/sirupsen/logrus"
-	"github.com/cloudradar-monitoring/cagent/win_perf_counters"
+	"github.com/cloudradar-monitoring/cagent/perfcounters"
 )
 
 const memGetTimeout = time.Second * 10
-var watcher = win_perf_counters.Watcher()
+var watcher = perfcounters.Watcher()
 
 func (ca *Cagent) MemResults() (MeasurementsMap, error) {
 	results := MeasurementsMap{}
--- a/mem_windows.go
+++ b/mem_windows.go
@@ -42,7 +42,6 @@ func (ca *Cagent) MemResults() (MeasurementsMap, error) {
 	}
 
 	free, err := watcher.Query(`\Memory\Free & Zero Page List Bytes`, "*")
-
 	if err != nil {
 		log.Errorf("[MEM] Failed to get free memory: %s", err.Error())
 	} else {
--- a/mem_windows.go
+++ b/mem_windows.go
@@ -1,4 +1,4 @@
-// +build !windows
+// +build windows
 
 package cagent
 
@@ -10,9 +10,11 @@ import (
 
 	"github.com/shirou/gopsutil/mem"
 	log "github.com/sirupsen/logrus"
+	"github.com/cloudradar-monitoring/cagent/win_perf_counters"
 )
 
 const memGetTimeout = time.Second * 10
+var watcher = win_perf_counters.Watcher()
 
 func (ca *Cagent) MemResults() (MeasurementsMap, error) {
 	results := MeasurementsMap{}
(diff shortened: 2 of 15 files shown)


soupdiver:
Please add a description about the changes you did, so I can get some context.
Opening a PR without _any_ information is not ideal.

mem.go
```diff
@@ -31,6 +33,7 @@ func (ca *Cagent) MemResults() (MeasurementsMap, error) {
 		results["buff_B"] = nil
 		results["available_B"] = nil
 	} else {
+
```
soupdiver: remove empty line

mem_windows.go
```diff
@@ -0,0 +1,57 @@
...
+		results["used_B"] = nil
+		results["available_B"] = nil
+	} else {
+		results["total_B"] = memStat.Total
+		results["used_B"] = memStat.Used
+		results["available_B"] = memStat.Available
+	}
+
+	free, err := watcher.Query(`\Memory\Free & Zero Page List Bytes`, "*")
+
```
soupdiver: remove empty line

soupdiver:
> We can't use it as a dependency because in their case it is an Input plugin for their own system. Maybe we can try to emulate telegraf but this will require a lot of messy abstraction interfaces to be done.

I see, yeah sounds messy to integrate that.
I have found https://github.com/lxn/win which I want to use instead of parsing the textual output of `typeperf`. To be hones I haven't understood that much yet about all the win32 api things and handlers etc.
The implementation looks similar to what you have included I think. Can you tell if this one is more abstract and maybe we can include this one instead our custom modifications?

requilence:
Actually `kernel32.go` and `pdh*.go` were taken by Telegraf from the https://github.com/lxn/win.

These files are not modified and license is included in the header

I think the reason this lib wasn't required as a dependency is that it is loading a ton of DLL's on the init(). In every file. E.g. https://github.com/lxn/win/blob/master/user32.go#L1668

soupdiver:
> Actually kernel32.go and pdh*.go were taken by Telegraf from the https://github.com/lxn/win.

All right, thanks for clarification.

> I think the reason this lib wasn't required as a dependency is that it is loading a ton of DLL's on the init(). 

Ah okay and I guess we also don't need that much library loading.

win_perf_counters/kernel32.go
```diff
@@ -0,0 +1,73 @@
...
+//
+// This is the official list of 'win' authors for copyright purposes.
+//
+// Alexander Neumann <an2048@googlemail.com>
+// Joseph Watson <jtwatson@linux-consulting.us>
+// Kevin Pors <krpors@gmail.com>
+
+// +build windows
+
+package win_perf_counters
```
soupdiver: Go package names should not contain underscores.
See. https://blog.golang.org/package-names

I would suggest to name the package `performancecounters`. Since I'm working on a implementation to retrieve CPU stats and will probably use the same approach as you did, I could extend that structure.

```
performancecounters/mem.go
performancecounters/mem_windows.go
performancecounters/kernel32.go
...
```

I can then add my `cpu.go`  and `cpu_windows.go`

requilence: Yes, I know about the package name style, it was just copied from the telegraf.
Renamed it. 
But I don't get what is the reason to put `mem.go`, `mem_windows.go` and `cpu*` inside the `perfcounters` folder? It

soupdiver: > Yes, I know about the package name style, it was just copied from the telegraf.

Glad we started with PRs then. If you know about the convention I wonder why you don't apply it then.

> But I don't get what is the reason to put mem.go, mem_windows.go and cpu* inside the perfcounters folder?

We group them in a package because their functionality is related. Retrieving CPU and memory info. Technically only windows implementation uses "real" performance counters but I think the package name still fits. Does that make sense or what you struggle to understand?

requilence: 1. Both `mem.go` and `cpu.go` includes a lot of Unix stuff 
2. If you move `mem_windows.go` and `cpu_windows.go` into `perfcounters` package directory, where do you plan to require `prefcounter` package?

soupdiver: > Both mem.go and cpu.go includes a lot of Unix stuff

I know. The windows implementation is in the `_windows.go` files. But from the point of view of our application we still have to put the logic somewhere together. I'm open for another name suggestion to avoid collision with the windows performance counters feature. I think `performancecounters` still fits our use case but to avoid confusion we can up with our own naming.

> If you move mem_windows.go and cpu_windows.go into perfcounters package directory, where do you plan to require prefcounter package?

I don't fully understand. The functionality to retrieve cpu/mem information is isolated in its own package and can be included from the `cagent` package. Maybe it's not really obvious at the moment because in the current structure everything is within one package, which is not a good practice. With putting this functionality in its own package we start to move to a more modular approach.

soupdiver: I just noticed that I was a step too far in my head already.
I suggested to put `mem.go` and the other files into the `performancecounters` package because I think that's a better way to organize our code.
But I don't wanna make this PR more complicated. Leave the `mem.go` and `mem_windows.go` where it is right now. I will add my new functionality to the `performancecounters` package then and at some point we can refactor the memory implementation.

Another think I noticed: Packages should be placed under the `pkg` package.
See: https://github.com/golang-standards/project-layout
`pkg/performancecounters` in our case. Sorry have overlooked that before

soupdiver: > Actually, I don't remember any lib that uses the pkg directory approach :)

We're building an application and not a standalone library. I agree that you don't necessarily need a `pkg` package in a library but when building an application it's good practice to put your own code there.

requilence: Yes, but ```Other projects will import these libraries expecting them to work, so think twice before you put something here :-)```
This looks like just a recommendation on how to `export` internal library to be used by other projects.

This is not wrong to just use the internal package naming

PR #9, opened 2018-10-21
Raid output changed

requilence:
**See DEV-369**

I made some changes.

Instead of `"raid.md0.array.state":"degraded"` I want to output `"raid.md0.degraded": "1"`
 Because:

- Clean state is useless info(means there were no pending writes when the array was shut down)
- As I told before this states are only returned by mdadm, not mdstat
- Mdadm can return multiple states.

Here is an example of output...
Mdstat:
```
Personalities : [raid6] [raid5] [raid4]
md0 : active raid5 sda1[0] sdd1[2] sdb1[1]
     1465151808 blocks level 5, 64k chunk, algorithm 2 [4/3] [UUU_]
    unused devices: <none>
```

Output:
```
{
  "md0.degraded": 1,
  "md0.physicaldevice.missing": 1,
  "md0.physicaldevice.state.sda1": "active",
  "md0.physicaldevice.state.sdb1": "active",
  "md0.physicaldevice.state.sdd1": "active",
  "md0.state": "active",
  "md0.type": "raid5"
}
```

More examples available in the `raid_test.go`

### DIFF
--- a/raid.go
+++ b/raid.go
@@ -20,9 +20,9 @@ type Raid struct {
 var failed = regexp.MustCompile("\\[([U_]+)\\]")
 
 func (r Raid) GetFailedAndMissingPhysicalDevices() (failedDevices []string, missingDevicesCount int) {
-	for _, fd := range r.Failed {
-		if fd < len(r.Devices) {
-			failedDevices = append(failedDevices, r.Devices[fd])
+	for _, deviceIndex := range r.Failed {
+		if deviceIndex < len(r.Devices) {
+			failedDevices = append(failedDevices, r.Devices[deviceIndex])
 		} else {
 			missingDevicesCount++
 		}
@@ -32,9 +32,9 @@ func (r Raid) GetFailedAndMissingPhysicalDevices() (failedDevices []string, miss
 
 func (r Raid) GetActivePhysicalDevices() []string {
 	var activeDevices []string
-	for _, fd := range r.Active {
-		if fd < len(r.Devices) {
-			activeDevices = append(activeDevices, r.Devices[fd])
+	for _, deviceIndex := range r.Active {
+		if deviceIndex < len(r.Devices) {
+			activeDevices = append(activeDevices, r.Devices[deviceIndex])
 		}
 	}
 	return activeDevices
@@ -73,6 +73,9 @@ func parseMdstat(data string) (RaidArrays, error) {
 		matches := failed.FindStringSubmatch(lines[n+1])
 
 		if len(matches) > 0 {
+			/// Parse raid array status from mdstat output e.g. "[UUU_]"
+			// if device is up("U") or down/missing ("_")
+
 			for i := 0; i < len(matches[1]); i++ {
 				if matches[1][i:i+1] == "_" {
 					raid.Failed = append(raid.Failed, i)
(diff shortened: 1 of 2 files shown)


raid.go
```diff
@@ -20,20 +14,30 @@ type Raid struct {
...
 	Failed  []int
+	Active  []int
 }
 
 var failed = regexp.MustCompile("\\[([U_]+)\\]")
 
-func (r Raid) GetFailedPhysicalDevices() []string {
-	devices := []string{}
+func (r Raid) GetFailedAndMissingPhysicalDevices() (failedDevices []string, missingDevicesCount int) {
 	for _, fd := range r.Failed {
```
soupdiver: what does `fd` stand for?
Maybe try to use more explicit variable names. Sometimes the short forms are obvious but I'm not sure in this case

raid.go
```diff
@@ -20,20 +14,30 @@ type Raid struct {
...
-			devices = append(devices, fmt.Sprintf("missing_device[%d]", fd))
+			missingDevicesCount++
+		}
+	}
+	return failedDevices, missingDevicesCount
+}
+
+func (r Raid) GetActivePhysicalDevices() []string {
+	var activeDevices []string
+	for _, fd := range r.Active {
```
soupdiver: What does `fd` stand for?

raid.go
```diff
@@ -72,12 +76,11 @@ func parseMdstat(data string) (RaidArrays, error) {
 			for i := 0; i < len(matches[1]); i++ {
 				if matches[1][i:i+1] == "_" {
 					raid.Failed = append(raid.Failed, i)
+				} else if matches[1][i:i+1] == "U" {
```
soupdiver: Please add some comment what this check is doing

raid.go
```diff
@@ -91,8 +94,24 @@ func (ar RaidArrays) Measurements() MeasurementsMap {
 		results[raid.Name+".state"] = raid.State
 		results[raid.Name+".type"] = raid.Type
 
-		if raid.State == RaidStatusDegraded {
-			results[raid.Name+".failed"] = strings.Join(raid.GetFailedPhysicalDevices(), "; ")
+		failedDevices, missingCount := raid.GetFailedAndMissingPhysicalDevices()
+
+		if len(failedDevices) > 0 || missingCount > 0 {
```
soupdiver: Please add a short comment what is checked here and the implications

raid.go
```diff
@@ -91,8 +94,24 @@ func (ar RaidArrays) Measurements() MeasurementsMap {
...
+			results[raid.Name+".degraded"] = 1
+
+			results[raid.Name+".physicaldevice.missing"] = missingCount
+
+			for _, failedDevice := range failedDevices {
```
soupdiver: Sort of ok in this scenario but I suggest you try to keep the branches inside `if/else` shorter.
Don't nest `if/else` and loops. This makes the code hard to read and functions can grow very quickly.
Try to extract the code within the branches into functions. Makes it easier to test, read and refactor.

raid.go
requilence: done

raid.go
requilence: done

raid.go
requilence: changed

raid.go
requilence: changed

raid.go
requilence: In this case, we have the combined data source for all raid arrays. If we will split the outer loop(arrays) and inner loops it will require to reparse the original data or store it in-the-middle

PR #10, opened 2018-10-21
PROC: add parent_PID

requilence:
See DEV-427:

...
Include the parent pid of each process.

 ```
{
    "PID": 24,
    "Name": "oom_reaper",
    "State": "sleeping",
    "Cmdline": "",
    "parent_PID": 0
}
```

### DIFF
--- a/processes_notwindows.go
+++ b/processes_notwindows.go
@@ -76,7 +76,13 @@ func processesFromProc() ([]ProcStat, error) {
 			log.Errorf("Failed to read cmdline(%s): %s", stats[0], err.Error())
 		}
 
-		stat := ProcStat{PID: pid, Name: string(bytes.TrimRight(comm, "\n")), Cmdline: strings.Replace(string(bytes.TrimRight(cmdline, "\x00")), "\x00", " ", -1)}
+		ppid, err := strconv.Atoi(string(stats[4]))
+
+		if err != nil {
+			log.Errorf("Failed to convert PPID(%s) to int: %s", stats[4], err.Error())
+		}
+
+		stat := ProcStat{PID: pid, ParentPID: ppid, Name: string(bytes.TrimRight(comm, "\n")), Cmdline: strings.Replace(string(bytes.TrimRight(cmdline, "\x00")), "\x00", " ", -1)}
 
 		switch stats[2][0] {
 		case 'R':
@@ -128,7 +134,7 @@ func execPS() ([]byte, error) {
 		return nil, err
 	}
 
-	out, err := exec.Command(bin, "axwwo", "pid,state,command").Output()
+	out, err := exec.Command(bin, "axwwo", "pid,ppid,state,command").Output()
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: processes_notwindows.go [hunk 1]
```diff
@@ -76,7 +76,13 @@ func processesFromProc() ([]ProcStat, error) {
 			log.Errorf("Failed to read cmdline(%s): %s", stats[0], err.Error())
 		}
 
-		stat := ProcStat{PID: pid, Name: string(bytes.TrimRight(comm, "\n")), Cmdline: strings.Replace(string(bytes.TrimRight(cmdline, "\x00")), "\x00", " ", -1)}
+		ppid, err := strconv.Atoi(string(stats[4]))
+
```
soupdiver: remove empty line

processes_notwindows.go
```diff
@@ -76,7 +76,13 @@ func processesFromProc() ([]ProcStat, error) {
...
+		if err != nil {
+			log.Errorf("Failed to convert PPID(%s) to int: %s", stats[4], err.Error())
+		}
+
+		stat := ProcStat{PID: pid, ParentPID: ppid, Name: string(bytes.TrimRight(comm, "\n")), Cmdline: strings.Replace(string(bytes.TrimRight(cmdline, "\x00")), "\x00", " ", -1)}
```
soupdiver: To make it easier to read I would suggest to create `ProcState` in multiple lines and not one.

processes_notwindows.go
```diff
@@ -161,12 +167,18 @@ func processesFromPS() ([]ProcStat, error) {
...
 
-		last := strings.Join(parts[2:], " ")
+		last := strings.Join(parts[3:], " ")
 		fileBaseWithArgs := filepath.Base(last)
 		fileBaseParts := strings.Fields(fileBaseWithArgs)
 
-		stat := ProcStat{PID: pid, Name: fileBaseParts[0], Cmdline: last}
-		switch parts[1][0] {
+		ppid, err := strconv.Atoi(string(parts[1]))
+
```
soupdiver: remove empty line

processes_notwindows.go
requilence: Sometimes I put the empty line before the `if` block. Do you have a link to some convention on golang empty lines? I can't find any mention in the https://github.com/golang/go/wiki/CodeReviewComments

So then I will propose the PR to make all the empty lines consistent.

processes_notwindows.go
```diff
@@ -113,6 +113,7 @@ func readProcFile(filename string) ([]byte, error) {
 	data, err := ioutil.ReadFile(filename)
 	if err != nil {
 		if os.IsNotExist(err) {
+			// if file doesn't exists
 			return nil, nil
```
soupdiver: Do we ignore the error in this case or do we have to return the `err`?

processes_notwindows.go
soupdiver: There is not _everything_ mentioned in these guidelines.
I like to group code blocks logically. Function call together with error handling , declaration of variables and also `if` blocks.
Seems this comment was a mistake, the empty line is fine before the if block.
My bad.

processes_notwindows.go
requilence: Will clarify this in another PR

soupdiver: hmmm not ideal but ok. I will open an issue an assign to you so we don't forget about this.

PR #13, opened 2018-10-23
DO NOT MERGE small broken commit to see that golangci does

soupdiver:
golangci test

### DIFF
--- a/config.go
+++ b/config.go
@@ -1,18 +1,17 @@
 package cagent
 
 import (
+	"bytes"
+	"crypto/x509"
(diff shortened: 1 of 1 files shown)


config.go
```diff
@@ -65,6 +64,8 @@ func New() *Cagent {
 	var defaultLogPath string
 	var rootCertsPath string
 
+	ioutil.ReadFile("/etc/hosts")
```
golangcibot: Error return value of `ioutil.ReadFile` is not checked

PR #14, opened 2018-10-23
Memory: calculate percentage and available memory on some systems

requilence:
See DEV-366.
I also improved the `available` memory metric.
- On Linux/Windows it uses the value provided by OS 
- On some OS(OpenBSD, Freebsd, Darwin) it calculated it as `free+cached+buffered` by [gopsutil]( https://github.com/shirou/gopsutil/blob/master/mem)
- Otherwise available metric must be nil

### DIFF
--- a/mem.go
+++ b/mem.go
@@ -57,15 +57,22 @@ func (ca *Cagent) MemResults() (MeasurementsMap, error) {
 		results["buff_B"] = memStat.Buffers
 		results["buff_percent"] = floatToIntPercent(float64(memStat.Buffers) / float64(memStat.Total))
 
-		// linux has available memory metric – just use it
-		if memStat != nil && runtime.GOOS == "linux" {
-			results["available_B"] = int(memStat.Available)
-		} else {
-			// otherwise calculate as (free + cached + buffered)
-			// assuming that OS will give up buffered and cached memory to other apps that need it if necessary
-			results["available_B"] = int(memStat.Free + memStat.Cached + memStat.Buffers)
+		var hasAvailableMememory bool
+		// linux has native MemAvailable metric since 3.14 kernel
+		// others calculated within github.com/shirou/gopsutil
+		switch runtime.GOOS {
+		case "linux", "freebsd", "openbsd", "darwin":
+			hasAvailableMetric = true
+		default:
+			hasAvailableMememory = false
 		}
-		results["available_percent"] = floatToIntPercent(float64(results["available_B"].(int)) / float64(memStat.Total))
+
+		if memStat != nil && hasAvailableMememory {
+			results["available_B"] = int(memStat.Available)
+			results["available_percent"] = floatToIntPercent(float64(results["available_B"].(int)) / float64(memStat.Total))
+		} else {
+			results["available_B"] = nil
+		}
 	}
 
 	if len(errs) == 0 {
--- a/mem.go
+++ b/mem.go
@@ -70,8 +70,6 @@ func (ca *Cagent) MemResults() (MeasurementsMap, error) {
 		if memStat != nil && hasAvailableMememory {
 			results["available_B"] = int(memStat.Available)
 			results["available_percent"] = floatToIntPercent(float64(results["available_B"].(int)) / float64(memStat.Total))
-		} else {
-			results["available_B"] = nil
 		}
 	}
 
--- a/mem.go
+++ b/mem.go
@@ -7,6 +7,7 @@ import (
 	"errors"
 	"strings"
 	"time"
+	"runtime"
 
 	"github.com/shirou/gopsutil/mem"
 	log "github.com/sirupsen/logrus"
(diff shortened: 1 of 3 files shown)


mem.go
```diff
@@ -23,22 +24,53 @@ func (ca *Cagent) MemResults() (MeasurementsMap, error) {
 
 	memStat, err := mem.VirtualMemoryWithContext(ctx)
 
+	results = map[string]interface{}{
```
soupdiver: For consistency I would stick with either capital or small letters in the fields. To avoid any maybe case-sensitive problems I would stick with small letters.
Also I would write out both or abbreviate both of the words. Not mix it.
If I assume correctly `B` is for bytes my preference would be `total_bytes` etc.

soupdiver: You put the declaration of `results` between `memStat, err := mem.VirtualMemoryWithContext(ctx)` and the corresponding error handling. Please fix the order

mem.go
```diff
@@ -23,22 +24,53 @@ func (ca *Cagent) MemResults() (MeasurementsMap, error) {
...
+			hasAvailableMemory = false
+		}
+
+		if memStat != nil && hasAvailableMemory {
+			results["available_B"] = int(memStat.Available)
+			results["available_percent"] = floatToIntPercent(float64(results["available_B"].(int)) / float64(memStat.Total))
+		}
 	}
 
 	if len(errs) == 0 {
```
soupdiver: It seems there is no way that we can have multiple errors. So this can be simplified

mem.go
```diff
@@ -23,22 +24,53 @@ func (ca *Cagent) MemResults() (MeasurementsMap, error) {
...
+		"shared_percent":    nil,
+		"buff_B":            nil,
+		"buff_percent":      nil,
+		"used_B":            nil,
+		"used_percent":      nil,
+		"available_B":       nil,
+		"available_percent": nil,
+	}
+
 	if err != nil {
```
soupdiver: Please fix this `if` branch.
Return early in error case and then you can omit the whole `else` branch.
```suggestion
	if err != nil {
		log.Errorf("[MEM] Failed to get virtual memory stat: %s", err.Error())
		return nil, errors.New("MEM: " + err.Error())
	}
	results["total_B"] = memStat.Total
	...
```

mem_windows.go
```diff
@@ -25,27 +25,43 @@ func (ca *Cagent) MemResults() (MeasurementsMap, error) {
 
 	memStat, err := mem.VirtualMemoryWithContext(ctx)
 
-	results["free_B"] = nil
-	results["shared_B"] = nil
-	results["buff_B"] = nil
+	results = map[string]interface{}{
+		"total_B":           nil,
```
soupdiver: same comment for the keys as before

mem_windows.go
```diff
@@ -25,27 +25,43 @@ func (ca *Cagent) MemResults() (MeasurementsMap, error) {
...
+		"free_percent":      nil,
+		"cached_B":          nil,
+		"cached_percent":    nil,
+		"used_B":            nil,
+		"used_percent":      nil,
+		"available_B":       nil,
+		"available_percent": nil,
+	}
 
 	if err != nil {
```
soupdiver: Same as before, please return early in error case and omit `else` branch.

mem_windows.go
```diff
@@ -25,27 +25,43 @@ func (ca *Cagent) MemResults() (MeasurementsMap, error) {
...
+
+	cached, err := watcher.Query(`\Memory\Cache Bytes`, "*")
+	if err != nil {
+		log.Errorf("[MEM] Failed to get cached memory: %s", err.Error())
+	} else {
+		results["cached_B"] = int(cached)
+		results["cached_percent"] = floatToIntPercent(float64(cached) / float64(memStat.Total))
 	}
 
 	if len(errs) == 0 {
```
soupdiver: Same as before, it seems there can only occur one error in the function.
Also as a good practice:
Try to invert your flow. You check for the happy path and expect a failure in the end. Inverting this makes the code much easier to read.
In the end of the function you should have the successful return statement. Inside the function try to check for the error cases and then immediately return (if possible).

types.go
```diff
@@ -14,3 +14,7 @@ type Result struct {
 	Measurements MeasurementsMap `json:"measurements"`
 	Message      interface{}     `json:"message"`
 }
+
+func floatToIntPercent(f float64) int{
+	return int(f*100+0.5)
+}
```
soupdiver: I think it's good practice to save files with a newline at the end

types.go
```diff
@@ -14,3 +14,7 @@ type Result struct {
 	Measurements MeasurementsMap `json:"measurements"`
 	Message      interface{}     `json:"message"`
 }
+
+func floatToIntPercent(f float64) int{
+	return int(f*100+0.5)
```
soupdiver: How does this convert a float to a percentage 🤔

requilence: I needed to make a method because I use it a lot, and it becomes a mess.
Renamed to floatToIntPercentRoundUP

mem_windows.go
requilence: actually, it could be more than 1 error. Fixed

mem_windows.go
requilence: fixed

mem.go
soupdiver: So the backend expects the field be named `metric_B`? Ok, still don't like it but if it's requirement fine.

> The reason probably that B stands for bytes, not bits
I think you rarely report bits in these cases.

mem_windows.go
requilence: keys are provided by @thorstenkramm

mem.go
requilence: fixed

types.go
requilence: fixed

mem.go
requilence: > You put the declaration of `results` between `memStat, err := mem.VirtualMemoryWithContext(ctx)` and the corresponding error handling. Please fix the order

fixed

types.go
soupdiver: It's fine to make it a function I just wonder about the calculation but maybe I don't really understand the incoming value.

requilence: 0.987 -> 99
0.102 -> 10

mem_windows.go
requilence: >Try to invert your flow
I do this approach when possible. In this case, I needed to add multiple errors but still continue. I've just forgotten to append some of them 😬

PR #15, opened 2018-10-23
PROC: make errors more clear

### LINKED ISSUE #12: Explain why error is ignored
As discussed in PR an explanation should be added why the error is ignored.
See: https://github.com/cloudradar-monitoring/cagent/blob/master/processes_notwindows.go#L119

requilence:
closes #12

### DIFF
--- a/processes_notwindows.go
+++ b/processes_notwindows.go
@@ -17,6 +17,8 @@ import (
 	log "github.com/sirupsen/logrus"
 )
 
+var errorProcessTerminated = fmt.Errorf("Process was terminated")
+
 func processes() ([]ProcStat, error) {
 	if runtime.GOOS == "linux" {
 		return processesFromProc()
@@ -46,10 +48,9 @@ func processesFromProc() ([]ProcStat, error) {
 	for _, filename := range filenames {
 		data, err := readProcFile(filename)
 		if err != nil {
-			return nil, err
-		}
-
-		if data == nil {
+			if err != errorProcessTerminated {
+				log.Error("readProcFile error ", err.Error())
+			}
 			continue
 		}
 
@@ -62,16 +63,7 @@ func processesFromProc() ([]ProcStat, error) {
 		pid, err := strconv.Atoi(string(stats[0]))
 		if err != nil {
 			log.Errorf("Failed to convert PID(%s) to int: %s", stats[0], err.Error())
-		}
-
-		comm, err := readProcFile(getHostProc() + "/" + string(stats[0]) + "/comm")
-		if err != nil {
-			log.Errorf("Failed to read comm(%s): %s", stats[0], err.Error())
-		}
-
-		cmdline, err := readProcFile(getHostProc() + "/" + string(stats[0]) + "/cmdline")
-		if err != nil {
-			log.Errorf("Failed to read cmdline(%s): %s", stats[0], err.Error())
+			continue
 		}
 
 		ppid, err := strconv.Atoi(string(stats[4]))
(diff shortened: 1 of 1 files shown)


processes_notwindows.go
```diff
@@ -46,10 +48,9 @@ func processesFromProc() ([]ProcStat, error) {
 	for _, filename := range filenames {
 		data, err := readProcFile(filename)
 		if err != nil {
```
soupdiver: Is it ok to ignore this error case completely? Only if err is `errorProcessTerminated` we log it.

processes_notwindows.go
```diff
@@ -62,28 +63,28 @@ func processesFromProc() ([]ProcStat, error) {
...
-		cmdline, err := readProcFile(getHostProc() + "/" + string(stats[0]) + "/cmdline")
-		if err != nil {
-			log.Errorf("Failed to read cmdline(%s): %s", stats[0], err.Error())
-		}
+		stat := ProcStat{PID: pid, ParentPID: ppid}
 
-		ppid, err := strconv.Atoi(string(stats[4]))
-		if err != nil {
-			log.Errorf("Failed to convert PPID(%s) to int: %s", stats[4], err.Error())
+		comm, err := readProcFile(getHostProc() + "/" + string(stats[0]) + "/comm")
```
soupdiver: Maybe add a comment to explain this block a little

processes_notwindows.go
```diff
@@ -62,28 +63,28 @@ func processesFromProc() ([]ProcStat, error) {
...
+		} else if err == nil {
+			stat.Name = string(bytes.TrimRight(comm, "\n"))
 		}
 
-		stat := ProcStat{
-			PID:       pid,
-			ParentPID: ppid,
-			Name:      string(bytes.TrimRight(comm, "\n")),
-			Cmdline:   strings.Replace(string(bytes.TrimRight(cmdline, "\x00")), "\x00", " ", -1),
+		cmdline, err := readProcFile(getHostProc() + "/" + string(stats[0]) + "/cmdline")
```
soupdiver: Maybe add a comment to explain this block a little

processes_notwindows.go
```diff
@@ -115,15 +116,15 @@ func processesFromProc() ([]ProcStat, error) {
 func readProcFile(filename string) ([]byte, error) {
 	data, err := ioutil.ReadFile(filename)
 	if err != nil {
+		// if file doesn't exists it means that process was closed after we got the directory listing
```
soupdiver: I noticed this addresses: https://github.com/cloudradar-monitoring/cagent/issues/12
You can include things like `Fixes #12` or `Closes #12` in your commit message and then the issue is automatically referenced and closed when merging the PR.

processes_notwindows.go
requilence: we log if it NOT errorProcessTerminated. 
errorProcessTerminated means that process was terminated after we got the directory listing
This is totally OK

processes_notwindows.go
requilence: What exactly do you want to explain? 
In prev lines there was 		
`pid, err := strconv.Atoi(string(stats[0]))`
To avoid converting `pid` back to string I used stats[0]

processes_notwindows.go
soupdiver: It's not about the argument or converting the pid. I don't have the full context while reviewing. I just wonder what reading `.../comm` is for. I'm not sure for what `comm` is the abbreviation. Communication, command, etc.
Without a clear variable name and comment you have to know details about Linux what this is doing. Something like `Reading command information for pid` would be already sufficient.

soupdiver: ignored because of inactivity

processes_notwindows.go
soupdiver: ignored because of inactivity

soupdiver:
Merging without 100% approval but we have to move on.

PR #16, opened 2018-10-25
dev

soupdiver:


cmd/cagent/main.go
```diff
@@ -51,6 +51,8 @@ func main() {
 	ca := cagent.New()
 	ca.SetVersion(version)
 
+	ioutil.ReadFile("/etc/passwd")
```
golangcibot: Error return value of `ioutil.ReadFile` is not checked

PR #22, opened 2018-11-01
fix typo in filename x68 -> x86

soupdiver:
Fix typo


requilence:
Windows users more used to 32bit/64bit suffixes.

What do you think @thorstenkramm?

soupdiver:
So `Windows_64bit` and `Windows_32bit`?

requilence:
Yes, In case @thorstenkramm agree and there aren't any difficulties from the frontend side in order to change the link format

PR #23, opened 2018-11-01
[CPU][Windows] fix util measurements

requilence:
On windows, perCPU measurements returned as Percentage, not CPU time
see https://github.com/shirou/gopsutil/issues/600

- Made a new calculation for windows (using percent not time)
- Filtered out outdated measurements (in case there was some timeouts)
- Added some comments to make the code more clear

### DIFF
--- a/cpu.go
+++ b/cpu.go
@@ -180,6 +180,7 @@ func (tsa *TimeSeriesAverage) Percentage() (map[int]ValuesMap, error) {
 					// looks like some problem happen and we don't have enough(more than 1) measurements
 					// This could happen if all CPU queries for the last d minutes were timeouted
 					sum[d][key] = -1
+					continue
 				}
 
 				secondsSpentOnThisTypeOfLoad := lastVal - tsa.TimeSeries[keyInt].Values[key]
--- a/cpu.go
+++ b/cpu.go
@@ -10,6 +10,7 @@ import (
 	"strings"
 	"sync"
 	"time"
+	"math"
 
 	"github.com/shirou/gopsutil/cpu"
 	"github.com/shirou/gopsutil/load"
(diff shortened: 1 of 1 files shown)


cpu.go
```diff
@@ -137,7 +145,49 @@ func (tsa *TimeSeriesAverage) Percentage() (map[int]ValuesMap, error) {
...
+
+					valSum += t.Values[key]
+					count++
+				}
+
+				// found the average for all collected measurements for the last d minutes
+				sum[d][key] = roundUpWithPrecision(valSum/float64(count), 2)
+
+			} else {
+
```
soupdiver: remove empty line

cpu.go
```diff
@@ -137,7 +145,49 @@ func (tsa *TimeSeriesAverage) Percentage() (map[int]ValuesMap, error) {
...
+				for i := keyInt; i < len(tsa.TimeSeries); i++ {
+					// allow 3 seconds(0.05min) outage to include metrics query time
+					if time.Since(tsa.TimeSeries[i].Time).Minutes() <= float64(d)+0.05 {
+						keyInt = i
+						hasMetrics = true
+						break
+					}
+				}
+				if !hasMetrics || keyInt == len(tsa.TimeSeries)-1 {
+					// looks like some problem happen and we don't have enough(more than 1) measurements
```
soupdiver: Try to keep it consistent. Don't start one line with a small letter and the next one with a capital one

soupdiver:
A note from my side: For future PR It would be nice if you can add some summary about _what_ you have changed in the PR. Short explanation about the problem (as you did) followed by a short summary what you did and why to solve the problem.

The code in this PR is not complicated but I have to dive in a bit to understand the timeseries that are iterated over, why some measurements are skipped etc. With a short explanation that would be more clear from the beginning.

requilence:
>A note from my side: For future PR It would be nice if you can add some summary about what you have changed in the PR.

Ok! This time I put a lot of comments inside the code itself :) 
Next time I will mix all the things together 😃

PR #25, opened 2018-11-02
Set default log level to error

soupdiver:
Change the default log level `error` and print a warning if the user has set the log level to something we don't expect.


soupdiver:
@requilence You're right. Removed the line

PR #28, opened 2018-11-06
Upgrade gopsutil & remove lxn/win dep

requilence:
- Upgrade `gopsutil` dep in order to solve DEV-495
- Remove `github.com/lxn/win`. Nothing depends on it


Gopkg.toml
```diff
@@ -35,7 +35,7 @@
 
 [[constraint]]
   name = "github.com/shirou/gopsutil"
-  version = "2.18.7"
+  branch = "master"
```
soupdiver: Shouldn't we still pin to a specific version?
Does the latest release also fix our issue?
https://github.com/shirou/gopsutil/releases/tag/v2.18.10
2.18.7 it not the latest release

requilence: The latest version(2.18.10) doesn't contain the fix (https://github.com/shirou/gopsutil/compare/v2.18.10...master)
BTW I don't see the reason to pin to the specific version. 
- We have a lock file
- We include the vendor dir
- We upgrade deps manually&separately when we need to do it

soupdiver: Even if we do all these steps, from just looking at the lockfile it's unclear which version we expect for build. To have reproducible builds this is a must.
`master` or `latest` or similar is never a proper version definition.

But I see the need to upgrade even if the version is not released yet that contains the fix.
Can you point the dependency to the commit hash the includes the fix? Then we can check later and depend on the next proper releases version.

I know it's a bit picky and detailed but I prefer to pay attention to these details 😅

requilence: done

requilence: I got your point. But in case we have "lock" file I think we kind of guarantee that it works only with version locked in the lock file.

Sometimes, even x.y."minor" changes could break the code. So that's why the lock-file approach became the "production" default.

PR #31, opened 2018-11-07
Service install, make console msg more clear

requilence:
see DEV-486
- Note that cagent service is running
- Do not mention how to start

Next, we will need to add the HUB_URL check on install to ensure firewall doesn't block outgoing connection

### DIFF
--- a/cmd/cagent/main.go
+++ b/cmd/cagent/main.go
@@ -276,19 +276,21 @@ func main() {
 
 			if err != nil {
 				fmt.Printf("Already running\n")
+			} else {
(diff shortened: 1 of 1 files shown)


cmd/cagent/main.go
```diff
@@ -276,19 +276,21 @@ func main() {
 
 			if err != nil {
 				fmt.Printf("Already running\n")
```
soupdiver: should we maybe terminate if already running?

requilence: I think there is no need to terminate here. It will only show the how-to-stop message(it could be useful) and the logs location

PR #33, opened 2018-11-13
-t flag to test the HUB connection & creds

requilence:
See DEV-486
We need a way to test the HUB connection & creds.
This will be run as a postscript in pkg installers in order to provide fast feedback for the user. 
It can be useful when a user has firewall blocking outgoing connection or made a typo in the credentials

### DIFF
--- a/cmd/cagent/main.go
+++ b/cmd/cagent/main.go
@@ -113,6 +113,8 @@ func main() {
 		err := ca.TestHub()
 		if err != nil {
 			fmt.Printf("Cagent HUB test failed: %s\n", err.Error())
+			os.Exit(1)
+			return
 		}
 
 		fmt.Printf("HUB connection test succeed and credentials are correct!\n")
--- a/cmd/cagent/main.go
+++ b/cmd/cagent/main.go
@@ -77,6 +77,7 @@ func main() {
 
 	serviceUninstallPtr := flag.Bool("u", false, fmt.Sprintf("stop and uninstall the system service(%s)", systemManager.String()))
 	printConfigPtr := flag.Bool("p", false, "print the active config")
+	testConfigPtr := flag.Bool("t", false, "test the HUB config")
 	versionPtr := flag.Bool("version", false, "show the cagent version")
 
 	flag.Parse()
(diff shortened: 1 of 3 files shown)


requilence:
For Windows we can try to integrate this test on the last step of MSI install. But I'm not sure if it worth the efforts to implement it 😬

handler.go
```diff
@@ -51,10 +51,39 @@ func (ca *Cagent) initHubHttpClient() {
 	}
 }
 
+func (ca *Cagent) TestHub() error {
+	ca.initHubHttpClient()
+	req, err := http.NewRequest("HEAD", ca.HubURL, nil)
+
```

handler.go
```diff
@@ -51,10 +51,39 @@ func (ca *Cagent) initHubHttpClient() {
...
+		return err
+	}
+
+	req.Header.Add("User-Agent", ca.userAgent())
+
+	if ca.HubUser != "" {
+		req.SetBasicAuth(ca.HubUser, ca.HubPassword)
+	}
+	resp, err := ca.hubHttpClient.Do(req)
+
```

cmd/cagent/main.go
```diff
@@ -108,6 +109,16 @@ func main() {
 		return
 	}
 
+	if *testConfigPtr {
+		err := ca.TestHub()
+		if err != nil {
+			fmt.Printf("Cagent HUB test failed: %s\n", err.Error())
```
soupdiver: panic in error case or return a proper exit code?

requilence: the problem is that `cagent -t` called in the `dpkg install` postscript. If I use the non-zero return code the installation will possibly fail. I will test this now

handler.go
```diff
@@ -51,10 +51,41 @@ func (ca *Cagent) initHubHttpClient() {
...
+	req, err := http.NewRequest("HEAD", ca.HubURL, nil)
+	if err != nil {
+		return err
+	}
+
+	req.Header.Add("User-Agent", ca.userAgent())
+	if ca.HubUser != "" {
+		req.SetBasicAuth(ca.HubUser, ca.HubPassword)
+	}
+
```
golangcibot: File is not `goimports`-ed

PR #34, opened 2018-11-21
Refactor network.go

requilence:
Split big Results() method

gocyclo:
```
12 cagent (*netWatcher).fillCountersMeasurements network.go:136:1
12 cagent (*netWatcher).ExcludedInterfacesByNameMap network.go:88:1
10 cagent IPAddresses network.go:220:1
```

### DIFF
--- a/network.go
+++ b/network.go
@@ -113,7 +113,7 @@ func (nw *netWatcher) ExcludedInterfacesByNameMap(allInterfaces []utilnet.Interf
 				nw.constantlyExcludedInterfaceCache[netIf.Name] = false
 				continue
 			}
-			
+
 			excludedInterfaces[netIf.Name] = struct{}{}
 			log.Debugf("[NET] interface excluded: %s", netIf.Name)
 		}
@@ -202,7 +202,7 @@ func (nw *netWatcher) Results() (MeasurementsMap, error) {
 
 	interfaces, err := utilnet.Interfaces()
 	if err != nil {
-		log.Errorf("[NET] Failed to read interfaces: ", err.Error())
+		log.Errorf("[NET] Failed to read interfaces: %s", err.Error())
 		return nil, err
 	}
 
--- a/network.go
+++ b/network.go
@@ -2,7 +2,6 @@ package cagent
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"net"
 	"regexp"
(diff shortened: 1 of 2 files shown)


network.go
```diff
@@ -50,137 +40,181 @@ func (nw *netWatcher) Results() (MeasurementsMap, error) {
...
+			} else if nw.cagent.NetInterfaceExcludeDisconnected && isInterfaceDown(&netIf) {
+				// exclude DOWN interface for now
+				// lets cache it as false and then we will only check UP/DOWN status
+				nw.constantlyExcludedInterfaceCache[netIf.Name] = false
+			} else {
+				// interface is not excluded
+				nw.constantlyExcludedInterfaceCache[netIf.Name] = false
+				continue
+			}
+
```
golangcibot: File is not `goimports`-ed

network.go
```diff
@@ -13,34 +12,25 @@ import (
...
-	var errs []string
-	ctx, cancel := context.WithTimeout(context.Background(), fsGetNetInterfacesTimeout)
-	defer cancel()
-
-	interfaces, err := utilnet.Interfaces()
-
-	if err != nil {
-		log.Errorf("[NET] Failed to read interfaces: %s", err.Error())
-		errs = append(errs, err.Error())
+func (nw *netWatcher) InterfaceExcludeRegexCompiled() []*regexp.Regexp {
```
soupdiver: Please add a (GoDoc compatible) summary what the function does.

network.go
```diff
@@ -50,137 +40,181 @@ func (nw *netWatcher) Results() (MeasurementsMap, error) {
...
+func (nw *netWatcher) isInterfaceExcludedByRegexp(netIf *utilnet.InterfaceStat) bool {
+	for _, re := range nw.InterfaceExcludeRegexCompiled() {
+		if re.MatchString(netIf.Name) {
+			return true
+		}
+	}
+	return false
+}
+
+func (nw *netWatcher) ExcludedInterfacesByNameMap(allInterfaces []utilnet.InterfaceStat) map[string]struct{} {
```
soupdiver: Maybe remove the Map from the function name?
There is no other `ExcludedInterfacesByName...` I think and we don't have to repeat the return type

network.go
```diff
@@ -50,137 +40,181 @@ func (nw *netWatcher) Results() (MeasurementsMap, error) {
...
+			excludedInterfaces[netIf.Name] = struct{}{}
+			log.Debugf("[NET] interface excluded: %s", netIf.Name)
 		}
 	}
+	return excludedInterfaces
+}
 
-	ctx, _ = context.WithTimeout(context.Background(), fsGetUsageTimeout)
-	counters, err := utilnet.IOCountersWithContext(ctx, true)
+func (nw *netWatcher) fillEmptyMeasurements(results MeasurementsMap, interfaces []utilnet.InterfaceStat, excludedInterfacesByName map[string]struct{}) {
```
soupdiver: please add some comment that explain what and why is filled here. I only see `= nil` and this doesn't fill something in my understanding 🤔

network.go
requilence: The main reason for this function is to return the map(to be able to check interface by name). Because the list(slice) itself was already available.

soupdiver: I see, so it converts the slice into a map. Then maybe rename to `ConvertExcludedInterfacesfromSliceToMap` Also quite verbose but more clear what's happening.

requilence: if a function called Convert it should accept the slice to convert. But in this case it is filter interfaces and provide a map of filtered out interfaces... BTW `ExcludedInterfacesByName` is ok in fact, because it contains `ByName` suffix and the argument `allInterfaces` is clear enough

network.go
requilence: I made some comment, is it clear enough?

network.go
requilence: done. Is it clear enough?

network.go
soupdiver: Ah I see, it creates all the keys inside the map

requilence:
Should I write GODOC comments or the others methods? I go through the file and the other methods look more self-descriptive

network.go
soupdiver: > But in this case it is filter interfaces and provide a map of filtered out interfaces

I see. These are the details that are perfectly placed in a short comment in front of the function :)

soupdiver:
> Should I write GODOC comments or the others methods

Ideally yes. This way we can generate a nice looking documentation for our code without even looking at it. This is a nice thing to provide to other people who want to use our code _without_ reading the whole code

requilence:
Sure!
Before we agreed to do this in a seperate PR. 
But I can start to do in this PR

soupdiver:
> Before we agreed to do this in a seperate PR.

I see.
 I'm fine with both ways. Proceed as you prefer

requilence:
> I'm fine with both ways. Proceed as you prefer

I will create the separate PR as soon as I finish with pending refactoring

PR #35, opened 2018-11-23
[cagent & frontman] rework default config

troian:
Generate minimum valuable config if there is no present by specified -c (or default option)
Add comment tags to config fields while creating new

Signed-off-by: Artur Troian <troian.ap@gmail.com>


cmd/cagent/main.go
```diff
@@ -142,17 +124,47 @@ func main() {
...
+
+		if response == "y" || response == "yes" {
+			return true
+		} else if response == "n" || response == "no" {
+			return false
+		}
+	}
+}
+
+func main() {
```
golangcibot: cyclomatic complexity 54 of func `main` is high (> 25)

troian:
Well, tricky question as it supports commenting which get simple and nice output to generated config.
This PR works in following manner:
 - returned ca.New() already has default values set which properly passed and written to output stream (e.g. cagent -p) and uses new toml library.
 - ReadConfigFromFile reads from existing config with old implementation till pelletier/go-toml#252 fixed.

Gopkg.toml
```diff
@@ -35,8 +35,7 @@
 
 [[constraint]]
   name = "github.com/shirou/gopsutil"
-  branch = "master"
-  revision = "4254ce2081c0d7994dae589a4550a4ca659ebf94"
+  revision = "v2.18.10"
```
requilence: Why did you change this? The specified commit contains a very important fix for the Windows CPU measurement.
https://github.com/shirou/gopsutil/compare/v2.18.10...4254ce2081c0d7994dae589a4550a4ca659ebf94

soupdiver: true, but it seems the current syntax is not even valid at all

requilence: ok, I got it. 
> only one of "branch", "version" or "revision" can be specified

@troian can you please remove `branch = "master" `

troian:
Should be ok now.
Locally go build used library from GOPATH instead of vendor

config.go
```diff
@@ -13,45 +13,60 @@ import (
...
 
-	HubURL           string `toml:"hub_url"`
-	HubGzip          bool   `toml:"hub_gzip"` // enable gzip when sending results to the HUB
-	HubUser          string `toml:"hub_user"`
-	HubPassword      string `toml:"hub_password"`
-	HubProxy         string `toml:"hub_proxy"`
-	HubProxyUser     string `toml:"hub_proxy_user"`
-	HubProxyPassword string `toml:"hub_proxy_password"`
+	// not nesting MinValuableConfig here as it will created entry with name MinValuableConfig when marshaling
+	HubURL           string `toml:"hub_url" comment:"# HTTP proxy to use with HUB"`
```
requilence: wrong comment here

requilence:
I like the idea of minimum valuable config! This will allow us to change with a new version the default config params that wasn't changed manually by the user .

Can you please make fields consistent and add "default" comment when it was missed?
It will be also good to have a line break between decription and default. Is it possible to insert within the struct tag?

troian:
> Now it is working 👍
> 
> I like the idea of minimum valuable config! This will allow us to change with a new version the default config params that wasn't changed manually by the user .
> 
> Can you please make fields consistent and add "default" comment where it was missed?
> It will be also good to have a line break between description and default. Is it possible to insert within the struct tag?

Yep, it's possible

.gitignore
```diff
@@ -0,0 +1,2 @@
+.idea
+cmd/cagent/cagent
```
soupdiver: 1) Why is that file ignored? Never had this artefact I think
2) No new-line at the end

cmd/cagent/main.go
```diff
@@ -28,99 +28,81 @@ var (
...
-	var serviceInstallUserPtr *string
-	var serviceInstallPtr *bool
-	outputFilePtr := flag.String("o", "", "file to write the results")
-
-	cfgPathPtr := flag.String("c", cagent.DefaultCfgPath, "config file path")
-	logLevelPtr := flag.String("v", defaultLogLevel, "log level – overrides the level in config file (values \"error\",\"info\",\"debug\")")
-	systemManager := service.ChosenSystem()
-	daemonizeModePtr := flag.Bool("d", false, "daemonize – run the proccess in background")
-	oneRunOnlyModePtr := flag.Bool("r", false, "one run only – perform checks once and exit. Overwrites output file")
+var ca *cagent.Cagent
```
soupdiver: Why using a global variables here?

cmd/cagent/main.go
```diff
@@ -28,99 +28,81 @@ var (
...
+var systemManager service.System
+
+func init() {
```
soupdiver: Why have you moved so much logic into the `init` function?
I don't think things like reading config file or dumping config or setting a logger belongs in this function.

cmd/cagent/main.go
```diff
@@ -330,16 +347,16 @@ func main() {
...
 		ca.Run(output, interruptChan, true)
 		return
-	} else {
-		go func() {
-			ca.Run(output, interruptChan, false)
-			doneChan <- struct{}{}
-		}()
 	}
 
+	go func() {
```
soupdiver: Add a comment why this is run in a goroutine

cmd/cagent/options.go
```diff
@@ -0,0 +1,17 @@
+package main
+
+var opts struct {
+	// Slice of bool will append 'true' each time the option
```
soupdiver: Which slice of bool are you referring to?

config.go
```diff
@@ -61,39 +76,14 @@ type Cagent struct {
...
-		Interval: 90,
-
-		CPULoadDataGather: []string{"avg1"},
-		CPUUtilTypes:      []string{"user", "system", "idle", "iowait"},
-		CPUUtilDataGather: []string{"avg1"},
-
-		FSTypeInclude: []string{"ext3", "ext4", "xfs", "jfs", "ntfs", "btrfs", "hfs", "apfs", "fat32"},
-		FSMetrics:     []string{"free_B", "free_percent", "total_B"},
-
+		LogFile:                         defaultLogPath,
```
soupdiver: Is `defaultLogPath ` set anywhere?

requilence:
> Is default value applicable for hub_url, hub_user, hub_password? Looks like not

No default value for them. I'm talking about LogLevel, HubGzip, NetInterfaceExclude

soupdiver:
I'm not too sure about splitting the default values into different files and use during compile time.
Sure it works, but I _personally_ think this does not benefit _that_ much but makes the code way less obvious. What do you think @requilence?

requilence:
> I'm not too sure about splitting the default values into different files and use during compile time.
> Sure it works, but I _personally_ think this does not benefit _that_ much but makes the code way less obvious. What do you think @requilence?

I don't get what do you mean. Are you talking about `options.go` file?

cmd/cagent/options.go
```diff
@@ -0,0 +1,17 @@
...
+	// is encountered (can be set multiple times, like -vvv)
+	OutputFile         string `short:"o" long:"output" description:"file to write the results"`
+	ConfigPath         string `short:"c" long:"config" description:"config file path"`
+	Verbose            string `short:"v" long:"verbose" default:"error" description:"log level – overrides the level in config file (values \"error\",\"info\",\"debug\")"`
```
soupdiver: This is inconsistent. The field is called `Verbose` but everywhere else we refer to it as `LogLevel` and also the description calls it `log level`

config.go
```diff
@@ -176,27 +167,67 @@ func (ca *Cagent) ReadConfigFromFile(configFilePath string, createIfNotExists bo
...
+		} else {
+			log.Infof("generated minimum valuable config: %s", configFilePath)
+		}
+	} else if configExists == nil {
+		_, err = btoml.DecodeFile(configFilePath, &ca)
 		if err != nil {
 			return err
 		}
+
+		// todo(troian) github.com/pelletier/go-toml completely overrides default values even if they omitted in the file
```
soupdiver: Should commented out code and todo stay here?

soupdiver:
> > I'm not too sure about splitting the default values into different files and use during compile time.
> > Sure it works, but I _personally_ think this does not benefit _that_ much but makes the code way less obvious. What do you think @requilence?
> 
> I don't get what do you mean. Are you talking about `options.go` file?

I talk about `defaults_darwin.go`, `defaults_windows.go` and defaults_others.go

requilence:
Ah, I see. Agree. OS-specific files usually are using to include code that is not able to compile on other systems. If you just want to set the constant it's more clear to just check `runtime.GOOS`

soupdiver:
> Ah, I see. Agree. OS-specific files usually are using to include code that is not able to compile on other systems. If you just want to set the constant it's more clear to just check `runtime.GOOS`

Agree

troian:
Agree flags stuff was not in place, so removed.

.gitignore
troian: That happens when need build and check without creating run configuration.
can quickly type some like CAGENT_HUB_URL="http://url" ./cagent -c new.conf

soupdiver: I see.
I usually use `go run` in these cases.
`CAGENT_HUB_URL="http://url" go run main.go -c new.conf`

troian:
Closing in favor of https://github.com/cloudradar-monitoring/cagent/pull/37

troian:
Closing in favor of https://github.com/cloudradar-monitoring/cagent/pull/37

PR #36, opened 2018-11-24
refactor handler.go

requilence:
- Split Run() method into (Run, RunOnce, GetAllMeasurements, ReportMeasurements)
- Move watchers into the Cagent struct

gocyclo before:
```
32 cagent (*Cagent).Run handler.go:135:1
11 cagent (*Cagent).TestHub handler.go:54:1
8 cagent (*Cagent).PostResultsToHub handler.go:86:1
7 cagent (*Cagent).initHubHttpClient handler.go:22:1
```
after: 
```
14 cagent (*Cagent).GetAllMeasurements handler.go:133:1
11 cagent (*Cagent).TestHub handler.go:52:1
8 cagent (*Cagent).PostResultsToHub handler.go:84:1
7 cagent (*Cagent).initHubHttpClient handler.go:20:1
5 cagent (*Cagent).Run handler.go:259:1
```

Todo: 
- Move different watchers into /pkg/monitoring
- Create an interface(New, Results)
- Make the code less repetitive (init watcher, get measurements, fill results map with a prefix)

But I think it is a good idea to merge this PR first

### DIFF
--- a/cmd/cagent/main.go
+++ b/cmd/cagent/main.go
@@ -330,12 +330,22 @@ func main() {
 	interruptChan := make(chan struct{})
 	doneChan := make(chan struct{})
 
+	if ca.PidFile != "" && *oneRunOnlyModePtr && runtime.GOOS != "windows" {
+		err := ioutil.WriteFile(ca.PidFile, []byte(strconv.Itoa(os.Getpid())), 0664)
+		if err != nil {
+			log.Errorf("Failed to write pid file at: %s", ca.PidFile)
+		}
+	}
+
 	if *oneRunOnlyModePtr == true {
-		ca.Run(output, interruptChan, true)
+		err := ca.RunOnce(output)
+		if err != nil {
+			log.Fatal(err)
+		}
 		return
 	} else {
 		go func() {
-			ca.Run(output, interruptChan, false)
+			ca.Run(output, interruptChan)
 			doneChan <- struct{}{}
 		}()
 	}
@@ -382,7 +392,7 @@ func (sw *serviceWrapper) Start(s service.Service) error {
 	sw.InterruptChan = make(chan struct{})
 	sw.DoneChan = make(chan struct{})
 	go func() {
-		sw.Cagent.Run(nil, sw.InterruptChan, false)
+		sw.Cagent.Run(nil, sw.InterruptChan)
 		sw.DoneChan <- struct{}{}
 	}()
 
--- a/cmd/cagent/main.go
+++ b/cmd/cagent/main.go
@@ -343,12 +343,12 @@ func main() {
 			log.Fatal(err)
 		}
 		return
-	} else {
-		go func() {
-			ca.Run(output, interruptChan)
-			doneChan <- struct{}{}
-		}()
 	}
+
+	go func() {
+		ca.Run(output, interruptChan)
+		doneChan <- struct{}{}
+	}()
 
 	select {
 	case sig := <-sigc:
--- a/config.go
+++ b/config.go
@@ -56,10 +56,10 @@ type Cagent struct {
 	// internal use
 	hubHttpClient *http.Client
 
-	cpuWatcher           *cpuWatcher
-	fsWatcher            *fsWatcher
-	netWatcher           *netWatcher
-	windowsUpdateWatcher *windowsUpdateWatcher
+	cpuWatcher           *CPUWatcher
+	fsWatcher            *FSWatcher
+	netWatcher           *NetWatcher
+	windowsUpdateWatcher *WindowsUpdateWatcher
 
 	rootCAs *x509.CertPool
 	version string
--- a/config.go
+++ b/config.go
@@ -56,8 +56,13 @@ type Cagent struct {
 	// internal use
 	hubHttpClient *http.Client
 
-	rootCAs *x509.CertPool
-	version string
+	cpuWatcher           *cpuWatcher
+	fsWatcher            *fsWatcher
+	netWatcher           *netWatcher
+	windowsUpdateWatcher *windowsUpdateWatcher
+	
+	rootCAs              *x509.CertPool
+	version              string
 }
 
 func New() *Cagent {
--- a/config.go
+++ b/config.go
@@ -60,9 +60,9 @@ type Cagent struct {
 	fsWatcher            *fsWatcher
 	netWatcher           *netWatcher
 	windowsUpdateWatcher *windowsUpdateWatcher
-	
-	rootCAs              *x509.CertPool
-	version              string
+
+	rootCAs *x509.CertPool
+	version string
 }
 
 func New() *Cagent {
--- a/cpu.go
+++ b/cpu.go
@@ -199,8 +199,8 @@ func (ca *Cagent) CPUWatcher() *cpuWatcher {
 		return ca.cpuWatcher
 	}
 
-	stat := cpuWatcher{}
-	stat.UtilAvg.mu.Lock()
+	cw := cpuWatcher{}
+	cw.UtilAvg.mu.Lock()
 
 	if len(ca.CPULoadDataGather) > 0 {
 		_, err := load.Avg()
@@ -214,11 +214,11 @@ func (ca *Cagent) CPUWatcher() *cpuWatcher {
 
 					switch v {
 					case 1:
-						stat.LoadAvg1 = true
+						cw.LoadAvg1 = true
 					case 5:
-						stat.LoadAvg5 = true
+						cw.LoadAvg5 = true
 					case 15:
-						stat.LoadAvg15 = true
+						cw.LoadAvg15 = true
 					default:
 						log.Errorf("[CPU] wrong cpu_load_data_gathering_mode. Supported values: avg1, avg5, avg15")
 					}
@@ -252,21 +252,21 @@ func (ca *Cagent) CPUWatcher() *cpuWatcher {
 		if !found {
 			log.Errorf("[CPU] utilisation metric '%s' not implemented on %s", t, runtime.GOOS)
 		} else {
-			stat.UtilTypes = append(stat.UtilTypes, t)
+			cw.UtilTypes = append(cw.UtilTypes, t)
 		}
 	}
 
-	stat.UtilAvg.SetDurationsMinutes(durations...)
-	stat.UtilAvg.mu.Unlock()
-	ca.cpuWatcher = &stat
+	cw.UtilAvg.SetDurationsMinutes(durations...)
+	cw.UtilAvg.mu.Unlock()
+	ca.cpuWatcher = &cw
 
+	// optimization to prevent CPU watcher to run in case CPU util metrics not are not needed
 	if len(ca.CPUUtilTypes) > 0 && len(ca.CPUUtilDataGather) > 0 || len(ca.CPULoadDataGather) > 0 {
-		// optimization to prevent CPU watcher to run in case CPU util metrics not are not needed
-		err := stat.Once()
+		err := cw.Once()
 		if err != nil {
 			log.Error("[CPU] Failed to read utilisation metrics: " + err.Error())
 		} else {
-			go stat.Run()
+			go cw.Run()
 		}
 	}
 
--- a/cpu.go
+++ b/cpu.go
@@ -44,7 +44,7 @@ type TimeSeriesAverage struct {
 	_DurationInMinutes []int // do not set directly, use SetDurationsMinutes
 }
 
-type CPUWatcher struct {
+type cpuWatcher struct {
 	LoadAvg1  bool
 	LoadAvg5  bool
 	LoadAvg15 bool
@@ -194,8 +194,12 @@ func (tsa *TimeSeriesAverage) Percentage() (map[int]ValuesMap, error) {
 	return sum, nil
 }
 
-func (ca *Cagent) CPUWatcher() *CPUWatcher {
-	stat := CPUWatcher{}
+func (ca *Cagent) CPUWatcher() *cpuWatcher {
+	if ca.cpuWatcher != nil {
+		return ca.cpuWatcher
+	}
+
+	stat := cpuWatcher{}
 	stat.UtilAvg.mu.Lock()
 
 	if len(ca.CPULoadDataGather) > 0 {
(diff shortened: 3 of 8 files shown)


cpu.go
```diff
@@ -194,8 +194,12 @@ func (tsa *TimeSeriesAverage) Percentage() (map[int]ValuesMap, error) {
 	return sum, nil
 }
 
-func (ca *Cagent) CPUWatcher() *CPUWatcher {
-	stat := CPUWatcher{}
+func (ca *Cagent) CPUWatcher() *cpuWatcher {
```
golangcibot: exported method CPUWatcher returns unexported type *cagent.cpuWatcher, which can be annoying to use

cpu.go
```diff
@@ -362,7 +377,7 @@ func (stat *CPUWatcher) Run() {
 	}
 }
 
-func (cs *CPUWatcher) Results() (MeasurementsMap, error) {
+func (cs *cpuWatcher) Results() (MeasurementsMap, error) {
```
golangcibot: receiver name cs should be consistent with previous receiver name stat for cpuWatcher

windows_fallback.go
```diff
@@ -2,12 +2,12 @@
 
 package cagent
 
-type EmptyWindowsUpdateWatcher struct{}
+type windowsUpdateWatcher struct{}
 
-func (ca *Cagent) WindowsUpdatesWatcher() EmptyWindowsUpdateWatcher {
-	return EmptyWindowsUpdateWatcher{}
+func (ca *Cagent) WindowsUpdatesWatcher() *windowsUpdateWatcher {
```
golangcibot: exported method WindowsUpdatesWatcher returns unexported type *cagent.windowsUpdateWatcher, which can be annoying to use

config.go
```diff
@@ -56,8 +56,13 @@ type Cagent struct {
 	// internal use
 	hubHttpClient *http.Client
 
-	rootCAs *x509.CertPool
-	version string
+	cpuWatcher           *cpuWatcher
+	fsWatcher            *fsWatcher
+	netWatcher           *netWatcher
+	windowsUpdateWatcher *windowsUpdateWatcher
+
```
golangcibot: File is not `goimports`-ed

config.go
```diff
@@ -56,8 +56,13 @@ type Cagent struct {
 	// internal use
 	hubHttpClient *http.Client
 
-	rootCAs *x509.CertPool
-	version string
+	cpuWatcher           *cpuWatcher
+	fsWatcher            *fsWatcher
+	netWatcher           *netWatcher
+	windowsUpdateWatcher *windowsUpdateWatcher
```
golangcibot: `windowsUpdateWatcher` is unused

cpu.go
requilence: will be solved when we move to separate pkgs and using interfaces

config.go
requilence: this is because you don't parse Windows files @golangcibot

config.go
requilence: fixed

cpu.go
requilence: fixed

windows_fallback.go
requilence: will be solved when we move to separate pkgs and using interfaces

cmd/cagent/main.go
```diff
@@ -330,12 +330,22 @@ func main() {
...
+	}
+
 	if *oneRunOnlyModePtr == true {
-		ca.Run(output, interruptChan, true)
+		err := ca.RunOnce(output)
+		if err != nil {
+			log.Fatal(err)
+		}
 		return
 	} else {
```
soupdiver: the `else` can be removed.
Inside the `if` branch either `log.Fatal` or `return` is called which end execution. Therefor the `else` can be omitted

config.go
soupdiver: > this is because you don't parse Windows files @golangcibot

Maybe open as issue at https://github.com/golangci/golangci and report

cpu.go
soupdiver: Is there a reason why the type can't or shouldn't be exported?

cpu.go
```diff
@@ -254,14 +258,25 @@ func (ca *Cagent) CPUWatcher() *CPUWatcher {
 
 	stat.UtilAvg.SetDurationsMinutes(durations...)
 	stat.UtilAvg.mu.Unlock()
+	ca.cpuWatcher = &stat
+
+	if len(ca.CPUUtilTypes) > 0 && len(ca.CPUUtilDataGather) > 0 || len(ca.CPULoadDataGather) > 0 {
```
soupdiver: Please comment what the if condition is.
If the explanation is the comment on line 264, please move it before the `if`

cpu.go
```diff
@@ -254,14 +258,25 @@ func (ca *Cagent) CPUWatcher() *CPUWatcher {
...
+		err := stat.Once()
+		if err != nil {
+			log.Error("[CPU] Failed to read utilisation metrics: " + err.Error())
+		} else {
+			go stat.Run()
+		}
+	}
 
-	return &stat
+	return ca.cpuWatcher
```
soupdiver: Since line 264 does set the value of ca.cpuWatcher we don't necessarily have to return the value again here I think. Except this makes using it easier in other places.
There are a couple of possible error cases, shouldn't we return the error in these cases?

fs.go
```diff
@@ -21,12 +21,16 @@ type fsWatcher struct {
 }
 
 func (ca *Cagent) FSWatcher() *fsWatcher {
-	fsWatcher := fsWatcher{AllowedTypes: map[string]struct{}{}, ExcludedPathCache: map[string]bool{}, cagent: ca}
+	if ca.fsWatcher != nil {
+		return ca.fsWatcher
+	}
+
+	ca.fsWatcher = &fsWatcher{AllowedTypes: map[string]struct{}{}, ExcludedPathCache: map[string]bool{}, cagent: ca}
```
soupdiver: Please use multi-line version to initialise the struct

handler.go
```diff
@@ -132,153 +130,137 @@ func (ca *Cagent) PostResultsToHub(result Result) error {
...
-
-		if err != nil {
-			log.Errorf("Failed to write pid file at: %s", ca.PidFile)
-		}
+	cpum, err := ca.CPUWatcher().Results()
+	if err != nil {
+		// no need to log because already done inside cpu.Results()
+		errs = append(errs, err.Error())
 	}
+	measurements = measurements.AddWithPrefix("cpu.", cpum)
```
soupdiver: please either have a new line before `measurements ...` or not

handler.go
```diff
@@ -132,153 +130,137 @@ func (ca *Cagent) PostResultsToHub(result Result) error {
...
-				// no need to log because already done inside RaidState()
-				errs = append(errs, err.Error())
-			}
+	return measurements, errors.New(strings.Join(errs, "; "))
+}
 
-			results.Measurements = results.Measurements.AddWithPrefix("raid.", raid)
-		}
+func (ca *Cagent) ReportMeasurements(measurements MeasurementsMap, outputFile *os.File) error {
+	result := Result{Timestamp: time.Now().Unix(), Measurements: measurements}
```
soupdiver: Please use multi-line

handler.go
```diff
@@ -132,153 +130,137 @@ func (ca *Cagent) PostResultsToHub(result Result) error {
...
 
-		if runtime.GOOS == "windows" {
-			wu, err := wuw.WindowsUpdates()
+	if outputFile != nil {
+		jsonEncoder := json.NewEncoder(outputFile)
+		err := jsonEncoder.Encode(result)
```
soupdiver: Don't you have to pass a reference here?

network.go
```diff
@@ -24,7 +24,12 @@ type netWatcher struct {
 }
 
 func (ca *Cagent) NetWatcher() *netWatcher {
-	return &netWatcher{cagent: ca, constantlyExcludedInterfaceCache: map[string]bool{}}
+	if ca.netWatcher != nil {
+		return ca.netWatcher
+	}
+
+	ca.netWatcher = &netWatcher{cagent: ca, constantlyExcludedInterfaceCache: map[string]bool{}}
```
soupdiver: Please use multi-line

windows.go
```diff
@@ -21,7 +21,7 @@ const (
 	WUStatusAborted
 )
 
-type WindowsUpdateWatcher struct {
+type windowsUpdateWatcher struct {
```
soupdiver: Why change to unexported?

cmd/cagent/main.go
requilence: thx, fixed

cpu.go
requilence: moved the comment

handler.go
```diff
@@ -132,153 +130,141 @@ func (ca *Cagent) PostResultsToHub(result Result) error {
...
-				errs = append(errs, err.Error())
-			}
+	return measurements, errors.New(strings.Join(errs, "; "))
+}
 
-			results.Measurements = results.Measurements.AddWithPrefix("raid.", raid)
-		}
+func (ca *Cagent) ReportMeasurements(measurements MeasurementsMap, outputFile *os.File) error {
+	result := Result{
+		Timestamp: time.Now().Unix(),
```

network.go
```diff
@@ -24,7 +24,15 @@ type netWatcher struct {
 }
 
 func (ca *Cagent) NetWatcher() *netWatcher {
-	return &netWatcher{cagent: ca, constantlyExcludedInterfaceCache: map[string]bool{}}
+	if ca.netWatcher != nil {
+		return ca.netWatcher
+	}
+
+	ca.netWatcher = &netWatcher{
+		cagent: ca,
```
golangcibot: File is not `goimports`-ed (from `goimports`)

windows.go
soupdiver: Yeah I get the idea. But @golangcibot is kinda right here, that it's annoying to use an unexported type. Code completion might not work and if the type is not exported it should only be used internally. So I would keep it exported.

If in some scenario you want to initialise the struct by yourself you're free to do so. Using the function is fine to be the intended way to get an instance of the struct. Maybe add a comment.

cpu.go
requilence: The idea is to have an instance of watcher inside the Cagent struct (`ca.cpuWatcher`). And `GetAllMeasurements()` should be able to retrieve the `cpuWatcher` instance using `ca.CPUWatcher()` in both case of cold run and after

handler.go
requilence: fixed

handler.go
requilence: fixed

handler.go
soupdiver: Seem you still don't have your run-goimports-on-save hook properly configured 😁

handler.go
requilence: not critical, but fixed

network.go
requilence: fixed

handler.go
soupdiver: Does it even work if you _don't_ pass a reference?
Usually things are passed by value, so they're copied. Therefor if you don't pass a reference the `Encode` function can't modify the struct you passed 🤔

network.go
requilence: actually, it is gofmted and goimported.. Tried multiple times

handler.go
requilence: fixed

cpu.go
soupdiver: I guess this answers my question `Except this makes using it easier in other places`.

What about the errors? Why don't we have to return them?

handler.go
requilence: it doesn't need to modify. It is encoder, not decoder :)

network.go
soupdiver: If I save the file, I get a change

soupdiver: screenshot 2018-11-26 at 11 43 43

handler.go
soupdiver: Gotcha... true 🙈

cpu.go
requilence: Cpu watcher has different types of measurements inside: different util, loadavg. Some of them could fail. If we return any of errors this will means that we lose the others. More, the error can self-heal after some time(e.g. WMI query on Windows can timeout on the first try)

Instead, we propose to return a watcher in any case. In case of error we just need to log them

soupdiver: ok, I see. Makes sense.
I had writing tests in mind... but while thinking about it, testing errors in code that measures CPU stuff might be (too) complicated anyway.
Ok then for the moment

network.go
requilence: I got the reason. IDE used Go v1.10.2. Looks like it doesn't add tabs in case there is a big difference. 
When I changed Go version to 1.11.2 it starts to format as you show 😅

windows.go
requilence: I rewrite this thing when I move all the watchers into separate pkgs. I just needed to refactor this file first so this is a very temp solution.

Can we merge it and I will propose moving PR that will have exported structs?

soupdiver: > Can we merge it and I will propose moving PR that will have exported structs?

yeah we can do this, I just wonder why we change it from exported to unexported and then change it back in another PR. Couldn't we just skip the hassle, leave it exported and still move into its own package in another PR? Or do I miss something?

requilence: because in case of fsWatcher and netWatcher they were already unexported. So I made it consistent. 
Then I realised that it will be changed to exported after we will move to pkgs

soupdiver: Aahhhh gotcha.
I would be fine either way but if we have to change something anyway what about export them all and be done with that?

requilence: Done 😁

config.go
```diff
@@ -56,6 +56,11 @@ type Cagent struct {
 	// internal use
 	hubHttpClient *http.Client
 
+	cpuWatcher           *CPUWatcher
+	fsWatcher            *FSWatcher
+	netWatcher           *NetWatcher
+	windowsUpdateWatcher *WindowsUpdateWatcher
```
golangcibot: `windowsUpdateWatcher` is unused (from `structcheck`)

fs.go
requilence: fixed

config.go
soupdiver: We ignore this then right @requilence ?

fs.go
```diff
@@ -14,22 +14,30 @@ import (
...
 
-func (ca *Cagent) FSWatcher() *fsWatcher {
-	fsWatcher := fsWatcher{AllowedTypes: map[string]struct{}{}, ExcludedPathCache: map[string]bool{}, cagent: ca}
+func (ca *Cagent) FSWatcher() *FSWatcher {
+	if ca.fsWatcher != nil {
+		return ca.fsWatcher
+	}
+
+	ca.fsWatcher = &FSWatcher{
+		AllowedTypes: map[string]struct{}{},
```
golangcibot: File is not `goimports`-ed (from `goimports`)

config.go
requilence: it is used only on windooows

soupdiver: yeah... seems @golangcibot is not checking for that case.

PR #37, opened 2018-11-26
[cagent & frontman] rework default config

troian:
Generate minimum valuable config if there is no present by specified -c (or default option)
Add comment tags to config fields while creating new

Add .gitignore
vendor: add github.com/pelletier/go-toml

Signed-off-by: Artur Troian <troian.ap@gmail.com>

includes https://github.com/cloudradar-monitoring/cagent/issues/38 as well


config.go
```diff
@@ -176,27 +187,62 @@ func (ca *Cagent) ReadConfigFromFile(configFilePath string, createIfNotExists bo
...
+		//  for now github.com/pelletier/go-toml completely overrides default values even if they omitted in the file
+
+		_, err = btoml.DecodeFile(configFilePath, &ca)
 		if err != nil {
 			return err
 		}
+
+		ca.lookupEnv()
+	} else {
+		return err
```
requilence: what error are you returning here?

requilence:
One small but uncool thing. This lib breaks the fields order:
```
darwin_amd64 git:(refactor-config-gen2) ✗ cat test2.conf        
# This is an auto-generated config to connect with the cloudradar service
# To see all options of cagent run cagent -p

# 
hub_password = ""

# HTTP url to connect with HUB
hub_url = ""

# 
hub_user = ""
```

In case of `cagent -p` this is total mess

config.go
troian: In case err is neither nil nor NotFound

soupdiver: This is another wild function.... so hard to read and follow the flow.

troian: Well, we can extend task to do refactoring

troian: Sounds good

config.go
```diff
@@ -13,45 +13,57 @@ import (
...
 	"time"
 
-	"github.com/BurntSushi/toml"
 	log "github.com/sirupsen/logrus"
+	"github.com/troian/toml"
 )
 
 var DefaultCfgPath string
 
+var configAutogeneratedHeadline = []byte(
```
golangcibot: `configAutogeneratedHeadline` is unused (from `deadcode`)

config.go
```diff
@@ -159,57 +173,90 @@ func (ca *Cagent) userAgent() string {
...
-			return fmt.Errorf("Failed to create the default config file: '%s'", configFilePath)
-		}
-		defer f.Close()
-		enc := toml.NewEncoder(f)
-		enc.Encode(ca)
-	} else {
-		_, err = os.Stat(configFilePath)
-		if err != nil {
-			return err
+	ca.lookupEnv()
```
soupdiver: The function is called `ReadConfigFromFile`. So it should not read something from environment. The name is misleading then

config.go
```diff
@@ -159,57 +173,90 @@ func (ca *Cagent) userAgent() string {
...
+
+	return err
+}
+
+func (ca *Cagent) CreateDefaultConfigFile(configFilePath string) error {
+	var err error
+	defer func() {
```
soupdiver: Is there a reason why this is done using `defer`?

config.go
```diff
@@ -159,57 +173,90 @@ func (ca *Cagent) userAgent() string {
...
+		if e := f.Close(); e != nil {
+			log.WithError(e).Errorf("failed to close config file: %s", configFilePath)
 		}
+	}()
+
+	if _, err = f.Write(configAutogeneratedHeadline); err != nil {
+		return fmt.Errorf("failed to write headline to config file")
 	}
 
+	ca.lookupEnv()
```
soupdiver: Same as before

config.go
```diff
@@ -159,57 +173,90 @@ func (ca *Cagent) userAgent() string {
...
+		return fmt.Errorf("config already exists at path: %s", configFilePath)
+	}
+
+	var f *os.File
+	if f, err = os.OpenFile(configFilePath, os.O_WRONLY|os.O_CREATE, 0644); err != nil {
+		return fmt.Errorf("failed to create the default config file: '%s'", configFilePath)
+	}
+
+	defer func() {
+		if e := f.Close(); e != nil {
```
soupdiver: Please use `err` as everywhere else

config.go
troian: Any type of error means there was issue so config may no be valid.

soupdiver: I get the error but I wonder _why_ you do this with defer?
Just check normally in the end of the function for an error

config.go
troian: Agree about rename but it should be here otherwise generated config contains empty values

config.go
troian: common approach is to to rewrite values with environment if they are provided
Might be function should be renamed to ```ReadConfig```

config.go
soupdiver: Why not call `lookupEnv()`  before calling this function then?
The more one function does internally the harder it becomes to write isolated tests later

config.go
soupdiver: Simply renaming does not solve the problem.
A function should (ideally) do _one_ thing. As said in the other comment, why not call `lookupEnv ()` before this function is called?

config.go
troian: Should caller take care of full process of config parse?

```go
if cfgPathPtr != nil {
		err := ca.ReadConfigFromFile(*cfgPathPtr)
		ca.ApplyEnv()
		if os.IsNotExist(err) {
			// this is ok
			err = ca.CreateDefaultConfigFile(*cfgPathPtr)
			if err != nil {
				log.Fatal(err)
			}
		} else if err != nil {
			if strings.Contains(err.Error(), "cannot load TOML value of type int64 into a Go float") {
				log.Fatalf("Config load error: please use numbers with a decimal point for numerical values")
			} else {
				log.Fatalf("Config load error: %s", err.Error())
			}
		}
	}
```

soupdiver: Looks better but pay attention to the logical flow.
Why would `ca.ApplyEnv()` only be called in the case of `if cfgPathPtr != nil`
Wouldn't it make more sense to call it _after_ the whole handling of config file is done? Meaning outside the `if` branch

troian: It was an example
Looks like ``` if cfgPathPtr != nil { ``` can be removed as by current flow it either provided by user or set as default path otherwise environment must be applied twice (if ```ReadConfigFromFile``` does not guarantee ca fields rewrite in case of error)

soupdiver: >Looks like if cfgPathPtr != nil { can be removed as by current flow it either provided by user or set as default path

true

So first try to read config from file and then override with values from env/flags in a defined order. We probably should document the order as well then.

config.go
```diff
@@ -13,45 +13,81 @@ import (
 	"strings"
 	"time"
 
-	"github.com/BurntSushi/toml"
 	log "github.com/sirupsen/logrus"
+	"github.com/troian/toml"
 )
 
+// Config instantiation must be done in multiple steps
```
soupdiver: Why is this here?

config.go
```diff
@@ -159,57 +195,83 @@ func (ca *Cagent) userAgent() string {
...
+	if _, err = f.Write(configAutogeneratedHeadline); err != nil {
+		return fmt.Errorf("failed to write headline to config file")
+	}
+
+	err = toml.NewEncoder(f).Encode(&MinValuableConfig{
+		HubURL:      ca.HubURL,
+		HubUser:     ca.HubUser,
+		HubPassword: ca.HubPassword,
+	})
+
```
soupdiver: remove empty line between assignment and error handling

cmd/cagent/main.go
```diff
@@ -93,17 +93,26 @@ func main() {
 
 	log.SetFormatter(&tfmt)
 
-	if cfgPathPtr != nil {
-		err := ca.ReadConfigFromFile(*cfgPathPtr, true)
+	err := ca.ReadConfigFromFile(*cfgPathPtr)
+	ca.ApplyEnv()
```
soupdiver: Move this after the error handling

config.go
soupdiver: A bunch of commented out code just causes confusion.
Add a small comment in the code where the initialisation is done to describe why it needs to be done in this order. Wrap all these steps in a function `HandleConfig` or similar.
I think there are better ways than having this big block of commented out code.

troian: Well, this is common approach for Go to provide an example of API usage.
That will look quite ugly

```go
ca := cagent.New()
ca.HandleConfig()
```

begs to

```go
func LoadConfig(file string) *Cagent {
	ca := New()

	err := ca.ReadConfigFromFile(file)
	ca.ApplyEnv()
	if os.IsNotExist(err) {
		// this is ok
		err = ca.CreateDefaultConfigFile(file)
		if err != nil {
			log.Fatal(err)
		}
	} else if err != nil {
		if strings.Contains(err.Error(), "cannot load TOML value of type int64 into a Go float") {
			log.Fatalf("Config load error: please use numbers with a decimal point for numerical values")
		} else {
			log.Fatalf("Config load error: %s", err.Error())
		}
	}

	if err = ca.Initialize(); err != nil {
		log.Fatal(err)
	}

	return ca
}
```

soupdiver: > Well, this is common approach for Go to provide an example of API usage.

Agree, but we don't document an API here for external developers. This is bootstrap code for the application and won't be used in any other place.

Another note to your example: The function is called `LoadConfig ` but also does instantiate `ca`. Functions that instantiate things internally are always hard to test. Create `ca` outside the function and pass it as an argument. Check google for `Dependency Injection`.

I image the whole config flow as following:
1. Try to read config file and set values in `ca`
2. If no config is present, try to create default config at location and load it. If this fails panic
3. Parse environment vars and overwrite values in `ca`
4. Check which flags are provided and overwrite values in `ca` in case

Each step can basically live in its own function and all steps can be combined in a helper function like `HandleConfig` or something similar named to keep `main` more clean.

troian: ok

config.go
```diff
@@ -165,7 +142,21 @@ func secToDuration(secs float64) time.Duration {
 	return time.Duration(int64(float64(time.Second) * secs))
 }
 
-// ApplyEnv read environment variables and replace respective fields if present
+func (ca *MinValuableConfig) ApplyEnv() {
```
soupdiver: use `mvc` but not `ca`

config.go
```diff
@@ -275,3 +265,27 @@ func (ca *Cagent) Initialize() error {
 
 	return nil
 }
+
+// HandleConfig instantiates Cagent with parameters specified in file
```
soupdiver: Comment is wrong. It does not `instantiates `

config.go
```diff
@@ -275,3 +265,27 @@ func (ca *Cagent) Initialize() error {
...
+// if config file not exists default one created in form of MinValuableConfig
+// todo: this function must be removed once config is separated from Cagent
+func HandleConfig(ca *Cagent, file string) error {
```
soupdiver: rename parameter `file` to `configFilePath`

PR #39, opened 2018-11-27
use proper windows paths

soupdiver:
Fix path for windows so signtool can find the certificate used for signing


requilence:
How it was working before :)?

PR #46, opened 2018-12-14
List hyperv vms

troian:
Include list of running Hyper-V VMs.
Output includes `virt.hyper-v` entry with with list of VMs

### DIFF
--- a/cagent.go
+++ b/cagent.go
@@ -70,10 +70,10 @@ func New(cfg *Config, version string) *Cagent {
 	for _, name := range ca.Config.VirtualMachinesStat {
 		vm, err := vmstat.Acquire(name)
 		if err != nil {
-			log.WithError(err).Warnf("acquire vm provider \"%s\"", name)
+			log.Warnf("acquire vm provider \"%s\": %s", name, err.Error())
 		} else {
 			if err = vm.IsAvailable(); err != nil {
-				log.WithError(err).Warnf("vm provider \"%s\" either not available or not enabled for host system %s", name, runtime.GOOS)
+				log.Warnf("vm provider \"%s\" either not available or not enabled for host system %s: %s", name, runtime.GOOS, err.Error())
 			}
 
 			ca.vmWatchers[name] = vm
@@ -99,7 +99,7 @@ func (ca *Cagent) userAgent() string {
 func (ca *Cagent) Shutdown() error {
 	for name, p := range ca.vmWatchers {
 		if err := vmstat.Release(p); err != nil {
-			log.WithError(err).Errorf("release vm provider \"%s\"", name)
+			log.Errorf("release vm provider \"%s\": %s", name, err.Error())
 		}
 
 		delete(ca.vmWatchers, name)
--- a/cagent.go
+++ b/cagent.go
@@ -9,6 +9,8 @@ import (
 	"runtime"
 	"strings"
 
+	"github.com/cloudradar-monitoring/cagent/pkg/monitoring/vmstat"
+
 	log "github.com/sirupsen/logrus"
 )
 
@@ -22,6 +24,7 @@ type Cagent struct {
 	fsWatcher            *FSWatcher
 	netWatcher           *NetWatcher
 	windowsUpdateWatcher *WindowsUpdateWatcher // nolint: structcheck,megacheck
+	vmWatchers           map[string]vmstat.Provider
 
 	rootCAs *x509.CertPool
 
@@ -30,8 +33,9 @@ type Cagent struct {
 
 func New(cfg *Config, version string) *Cagent {
 	ca := &Cagent{
-		Config:  cfg,
-		version: version,
+		Config:     cfg,
+		version:    version,
+		vmWatchers: make(map[string]vmstat.Provider),
 	}
 
 	if rootCertsPath != "" {
@@ -59,6 +63,19 @@ func New(cfg *Config, version string) *Cagent {
 		}
 	}
 
+	for _, name := range ca.Config.VirtualMachinesStat {
+		vm, err := vmstat.Acquire(name)
+		if err != nil {
+			log.WithError(err).Warnf("acquire vm provider \"%s\"", name)
+		} else {
+			if err = vm.IsAvailable(); err != nil {
+				log.WithError(err).Warnf("vm provider \"%s\" either not available or not enabled for host system %s", name, runtime.GOOS)
+			}
+
+			ca.vmWatchers[name] = vm
+		}
+	}
+
 	return ca
 }
 
@@ -73,4 +90,16 @@ func (ca *Cagent) userAgent() string {
 	parts := strings.Split(ca.version, "-")
 
 	return fmt.Sprintf("Cagent v%s %s %s", parts[0], runtime.GOOS, runtime.GOARCH)
+}
+
+func (ca *Cagent) Shutdown() error {
+	for name, p := range ca.vmWatchers {
+		if err := vmstat.Release(p); err != nil {
+			log.WithError(err).Errorf("release vm provider \"%s\"", name)
+		}
+
+		delete(ca.vmWatchers, name)
+	}
+
+	return nil
 }
--- a/cagent.go
+++ b/cagent.go
@@ -63,6 +63,10 @@ func New(cfg *Config, version string) *Cagent {
 		}
 	}
 
+	if err := InitVMStat(); err != nil {
+		log.Error("cannot instantiate virtual machines API: ", err.Error())
+	}
+
 	for _, name := range ca.Config.VirtualMachinesStat {
 		vm, err := vmstat.Acquire(name)
 		if err != nil {
--- a/pkg/monitoring/top/top_windows.go
+++ b/pkg/monitoring/top/top_windows.go
@@ -6,17 +6,15 @@ import (
 	"log"
 	"time"
 
-	"github.com/cloudradar-monitoring/cagent/perfcounters"
+	"github.com/cloudradar-monitoring/cagent/pkg/monitoring"
+
 	"github.com/pkg/errors"
 )
 
-var watcher *perfcounters.WinPerfCountersWatcher
 var counterPath = "\\Process(*)\\% Processor Time"
 
 func (t *Top) startCollect(interval time.Duration) {
-	watcher = perfcounters.Watcher()
-
-	err := watcher.StartContinousQuery(counterPath, interval)
+	err := monitoring.GetWatcher().StartContinuousQuery(counterPath, interval)
 	if err != nil {
 		log.Printf("Failed to StartQuery: %s", err)
 		return
@@ -24,7 +22,7 @@ func (t *Top) startCollect(interval time.Duration) {
 }
 
 func (t *Top) GetProcesses() ([]*ProcessInfo, error) {
-	res, err := watcher.GetFormattedQueryData(counterPath)
+	res, err := monitoring.GetWatcher().GetFormattedQueryData(counterPath)
 	if err != nil {
 		return nil, errors.Wrap(err, "Failed to call GetFormattedQueryData")
 	}
--- a/pkg/monitoring/vmstat/hyperv/hyperv_wmi.go
+++ b/pkg/monitoring/vmstat/hyperv/hyperv_wmi.go
@@ -3,7 +3,6 @@
 package hyperv
 
 import (
-	"fmt"
 	"net"
 	"regexp"
 	"strconv"
@@ -49,6 +48,9 @@ type msvm_ComputerSystem struct {
 	ProcessID             uint32
 }
 
+// some of fields are pointers instead of values as in some cases they may not exists
+// during WMI query. For example if VM is not running MemoryUsage, ProcessorLoad and Heartbeat
+// are not available
 type msvm_SummaryInformation struct {
 	Name                 string
 	ElementName          string
@@ -57,11 +59,12 @@ type msvm_SummaryInformation struct {
 	EnabledState         EnabledState
 	HealthState          HealthState
 	CreationTime         time.Time
-	MemoryUsage          *uint64
-	Uptime               uint64
-	ProcessorLoad        *uint16
-	Heartbeat            *Heartbeat
-	NumberOfProcessors   uint16
+	// MemoryUsage is in megabytes
+	MemoryUsage        *uint64
+	Uptime             uint64
+	ProcessorLoad      *uint16
+	Heartbeat          *Heartbeat
+	NumberOfProcessors uint16
 }
 
 type msvm_GuestNetworkAdapterConfiguration struct {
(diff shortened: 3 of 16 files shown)


pkg/monitoring/vmstat/vmstat.go
```diff
@@ -0,0 +1,88 @@
...
+}
+
+func Acquire(name string) (Provider, error) {
+	lock.Lock()
+	defer lock.Unlock()
+
+	if entry, ok := providers[name]; ok {
+		entry.wg.Add(1)
+		entry.run.Do(func() {
+			entry.prov.Run()
```
golangcibot: Error return value of `entry.prov.Run` is not checked (from `errcheck`)

troian: done

pkg/monitoring/vmstat/hyperv/hyperv_wmi.go
```diff
@@ -0,0 +1,228 @@
...
+	stat := &hypervStat{
+		List: make(map[string]vmStat),
+	}
+
+	var countersErr error
+	res, countersErr := im.watcher.GetFormattedQueryData(hypervPath)
+	if countersErr == nil {
+		// Iterate over the reported counter values
+		for _, c := range res {
+			// Some filtering...
```
soupdiver: If a comment is added, make it more useful :)

pkg/monitoring/vmstat/hyperv/hyperv_wmi.go
```diff
@@ -0,0 +1,228 @@
...
+			default:
+				name := strings.Split(c.InstanceName, ":")[0]
+				stat.List[name] = vmStat{
+					InstanceName:           name,
+					CPUWaitTimePerDispatch: c.Value,
+				}
+			}
+		}
+	} else {
+		fmt.Println(countersErr.Error())
```
soupdiver: `fmt` and not `log`?

pkg/monitoring/vmstat/hyperv/hyperv_wmi.go
```diff
@@ -0,0 +1,228 @@
...
+				vmEntry["cpu_wait_time_per_dispatch_ns"] = int64(stat.List[dst[i].ElementName].CPUWaitTimePerDispatch)
+			}
+			vmEntry["name"] = dst[i].ElementName
+			vmEntry["guid"] = dst[i].Name
+			vmEntry["guest_operating_system"] = dst[i].GuestOperatingSystem
+			vmEntry["version"] = dst[i].Version
+			vmEntry["enabled_state"] = dst[i].EnabledState.String()
+			vmEntry["health_state"] = dst[i].HealthState.String()
+			vmEntry["creation_time"] = dst[i].CreationTime
+			// dst[i].MemoryUsage is in megabytes
```
soupdiver: do we have to keep that line?

pkg/monitoring/vmstat/hyperv/provider_windows.go
```diff
@@ -0,0 +1,70 @@
...
+
+	log "github.com/sirupsen/logrus"
+)
+
+func init() {
+	prov := newImpl()
+
+	err := vmstat.RegisterVMProvider(prov)
+	if err != nil {
+		panic(err.Error())
```
soupdiver: should we really panic?

pkg/monitoring/vmstat/hyperv/provider_windows.go
```diff
@@ -0,0 +1,70 @@
...
+		watcher: monitoring.GetWatcher(),
+	}
+}
+
+func (im *impl) Run() error {
+	interval := time.Second * 1
+
+	err := im.watcher.StartContinuousQuery(hypervPath, interval)
+	if err != nil {
+		log.Printf("Failed to StartQuery: %s", err)
```
soupdiver: In case we return the `err` we should log the error in the level above.
If an error is logged it's basically handled.

pkg/monitoring/vmstat/onbuild_dependency/windows.go
```diff
@@ -0,0 +1,7 @@
+// +build windows
```
soupdiver: purpose of this file?

pkg/monitoring/vmstat/vmstat.go
```diff
@@ -0,0 +1,98 @@
...
+	GetMeasurements() (map[string]interface{}, error)
+}
+
+type provEntry struct {
+	prov Provider
+	wg   sync.WaitGroup
+	run  sync.Once
+}
+
+var providers map[string]*provEntry
```
soupdiver: Do we really need global variables?
Usually it's better and cleaner and less error prone if you don't have globals.
Can we crate a type, put these in there and then operate on a type directly?

pkg/monitoring/watcher_windows.go
```diff
@@ -0,0 +1,17 @@
+// +build windows
+
+package monitoring
+
+import (
+	"github.com/cloudradar-monitoring/cagent/perfcounters"
+)
+
+var watcher *perfcounters.WinPerfCountersWatcher
```
soupdiver: Again, can we avoid globals?

virt_windows.go
```diff
@@ -0,0 +1,7 @@
+// +build windows
```
soupdiver: purpose of this file?

pkg/monitoring/vmstat/hyperv/hyperv_wmi.go
troian: done

pkg/monitoring/vmstat/hyperv/hyperv_wmi.go
troian: done

pkg/monitoring/vmstat/hyperv/hyperv_wmi.go
troian: Hm, yeah, struct comment sounds better

pkg/monitoring/vmstat/hyperv/provider_windows.go
troian: done

pkg/monitoring/vmstat/onbuild_dependency/windows.go
troian: To fix cyclic dependencies

virt_windows.go
troian: To fix cyclic dependencies

pkg/monitoring/vmstat/vmstat.go
troian: We can do something like, but that again comes to that variable. for that case it's not like a global, more like a `static` analog to C++. Issue is register happens on `init` stage and all that is necessary as some VM implementations are OS specific and we should not allow `*nix` builds to acquire VM provider specific to windows only
```go
type VMStat struct {
}

var vmstat *vmStat
```

pkg/monitoring/watcher_windows.go
troian: It can be done but probably not within given PR. Some few other objects need this and to do it properly `Cagent` must allocate it and pass during `New` or `Configure` for `Top` `VM` etc.
But that will need to hide context of watcher type to `interface` as only windows stuff uses windows perfcounters

pkg/monitoring/vmstat/hyperv/provider_windows.go
troian: I think so as it happens at init stage and doing some logging is not proper here. This panic is more for develop and test purpose.
And to prevent duplicated entries in config (which may be cause of panic because such vm provider already exists) we should do config validation after reading file

soupdiver: > This panic is more for develop and test purpose.
This sounds odd.

Still, causing panic inside a package is no good practice. The error should bubble up and `log.Fatal` or something similar should be called on a higher level then.

virt_windows.go
troian: It cannot be included neither `Cagent` not other files as it is windows specific dependency (in can be included actually into `processes_windows.go` if we don't want to have empty file)
Idea is VM provider is registered by `func init()`. This prevent instantiating inappropriate VM provider from within unsupported host system. To call that init we need somehow import VM provider package

soupdiver: I see... nasty OS dependency 😅 
Please add a comment why we have this empty file. You can then resolve this conversation

pkg/monitoring/vmstat/vmstat.go
troian: Ok. I'll pack
Concurrency already being handled

cagent.go
```diff
@@ -59,6 +63,23 @@ func New(cfg *Config, version string) *Cagent {
...
+		log.Error("cannot instantiate virtual machines API: ", err.Error())
+	}
+
+	for _, name := range ca.Config.VirtualMachinesStat {
+		vm, err := vmstat.Acquire(name)
+		if err != nil {
+			log.WithError(err).Warnf("acquire vm provider \"%s\"", name)
+		} else {
+			if err = vm.IsAvailable(); err != nil {
+				log.WithError(err).Warnf("vm provider \"%s\" either not available or not enabled for host system %s", name, runtime.GOOS)
```
requilence: we don't use `logrus.WithError` in other parts of code. IMO this is not for user-facing errors and more suitable for central logging systems like logstash.

Instead, we describe the context of error first and then append the original error after semicolon

cagent.go
```diff
@@ -74,3 +95,15 @@ func (ca *Cagent) userAgent() string {
 
 	return fmt.Sprintf("Cagent v%s %s %s", parts[0], runtime.GOOS, runtime.GOARCH)
 }
+
+func (ca *Cagent) Shutdown() error {
+	for name, p := range ca.vmWatchers {
+		if err := vmstat.Release(p); err != nil {
+			log.WithError(err).Errorf("release vm provider \"%s\"", name)
```
requilence: the same here (WithError)

pkg/monitoring/top/top_windows.go
```diff
@@ -6,25 +6,23 @@ import (
...
-var watcher *perfcounters.WinPerfCountersWatcher
 var counterPath = "\\Process(*)\\% Processor Time"
 
 func (t *Top) startCollect(interval time.Duration) {
-	watcher = perfcounters.Watcher()
-
-	err := watcher.StartContinousQuery(counterPath, interval)
+	err := monitoring.GetWatcher().StartContinuousQuery(counterPath, interval)
 	if err != nil {
 		log.Printf("Failed to StartQuery: %s", err)
```
requilence: Why not error?
Also, please can you prepend context, like `[VM]` because this is a global logging message. We did this way in the other parts of the code

pkg/monitoring/vmstat/hyperv/hyperv_wmi.go
```diff
@@ -0,0 +1,234 @@
...
+			if !guestNetworkRegexp.Match([]byte(ips[i].InstanceID)) {
+				log.Warn("invalid InstanceID in hyper-v response. expected form:"+
+					"[Microsoft:GuestNetwork\\<GUID>\\<GUID>] actual form: [%s]", ips[i].InstanceID)
+			} else {
+				id := guestNetworkRegexp.FindAllStringSubmatch(ips[i].InstanceID, -1)
+				ipMap[id[0][1]] = ips[i].IPAddresses
+			}
+		}
+	} else {
+		log.WithError(err)
```
requilence: unfinished error logging

pkg/monitoring/vmstat/hyperv/hyperv_wmi.go
```diff
@@ -0,0 +1,234 @@
...
+	}
+
+	var dst []msvm_SummaryInformation
+	q = wmi.CreateQuery(&dst, "")
+
+	if err := wmi.QueryNamespace(q, &dst, `root\virtualization\v2`); err != nil {
+		log.WithError(err)
```
requilence: the same here. `log.WithError(err)` doesn't log the error

pkg/monitoring/vmstat/hyperv/hyperv_wmi.go
```diff
@@ -0,0 +1,234 @@
...
+	case 2:
+		return "ok"
+	case 6:
+		return "error"
+	case 12:
+		return "no contact"
+	case 13:
+		return "lost communication"
+	default:
+		return "unknown"
```
requilence: in other String() methods default returns `<unknown>`

pkg/wmi/wmi.go
```diff
@@ -0,0 +1,55 @@
...
+func (st FeatureInstallState) String() string {
+	switch st {
+	case 1:
+		return "enabled"
+	case 2:
+		return "disabled"
+	case 3:
+		return "absent"
+	default:
+		return "unknown"
```
requilence: in other String() methods default returns <unknown>

pkg/monitoring/vmstat/hyperv/hyperv_wmi.go
```diff
@@ -0,0 +1,234 @@
...
+			default:
+				name := strings.Split(c.InstanceName, ":")[0]
+				stat.List[name] = vmStat{
+					InstanceName:           name,
+					CPUWaitTimePerDispatch: c.Value,
+				}
+			}
+		}
+	} else {
+		log.WithError(countersErr)
```
requilence: unfinished withError

cagent.go
soupdiver: good point, agree

pkg/monitoring/top/top_windows.go
requilence: Oh, sorry. This line wasn't made by you. Maybe @soupdiver can address it here on in separate PRs
Still, you have other logging where you could prepend the context ("[VM]")

cagent.go

cagent.go

pkg/monitoring/top/top_windows.go
troian: done

pkg/monitoring/vmstat/hyperv/hyperv_wmi.go
troian: done

pkg/monitoring/vmstat/hyperv/hyperv_wmi.go
troian: done

pkg/wmi/wmi.go
troian: done

pkg/monitoring/vmstat/hyperv/hyperv_wmi.go
troian: done

pkg/monitoring/vmstat/hyperv/hyperv_wmi.go
troian: done

PR #51, opened 2018-12-19
Fix processes info parsing

requilence:
improve both `/proc/<pid>` and ps methods.
Because our goal is to create the tool for a wide range of systems we should try to avoid relating to unclear structures
What changed in this PR:
- in the case of `ps` method parse the header to make sure ps returns requested fields(and in the right order)
- in case of `/proc/<pid>` method switch from `/proc/<pid>/stat` to the `/proc/<pid>/status` parsing. According to some sources `ppid` is 3rd, other shows it 4th so it may differ from system to system. Also, it turned out that 2nd param(executable name) can contain spaces and breaks the parsing

This should theoretically solve DEV-654 and other similar issues

### DIFF
--- a/processes_notwindows.go
+++ b/processes_notwindows.go
@@ -85,9 +85,6 @@ func processesFromProc() ([]ProcStat, error) {
 		stat := ProcStat{ParentPID: procStatus.PPID, State: procStatus.State}
 		// get the PID from the filepath(/proc/<pid>/status) itself
 		pathParts := strings.Split(statusFilepath, string(filepath.Separator))
-		fmt.Println(statusFilepath)
-		fmt.Printf("%+v", pathParts)
-
 		pidString := pathParts[len(pathParts)-2]
 		stat.PID, err = strconv.Atoi(pidString)
 		if err != nil {
--- a/processes_notwindows.go
+++ b/processes_notwindows.go
@@ -134,7 +134,6 @@ func parseProcStatusFile(b []byte) procStatus {
 			if err != nil {
 				log.Errorf("[PROC] proc/status: failed to convert PPID(%s) to int: %s", fields[1], err.Error())
 			}
-			break
 		case "state:":
 			// extract raw long state
 			// eg "State:	S (sleeping)"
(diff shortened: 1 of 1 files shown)


processes_notwindows.go
```diff
@@ -27,90 +33,132 @@ func processes() ([]ProcStat, error) {
...
+		fields := strings.Fields(scanner.Text())
+
+		switch strings.ToLower(fields[0]) {
+		case "ppid:":
+			var err error
+			status.PPID, err = strconv.Atoi(fields[1])
+			if err != nil {
+				log.Errorf("[PROC] proc/status: failed to convert PPID(%s) to int: %s", fields[1], err.Error())
+			}
+			break
```
golangcibot: redundant break statement (from `megacheck`)

processes_notwindows.go
```diff
@@ -27,90 +33,132 @@ func processes() ([]ProcStat, error) {
...
+			// eg "State:	S (sleeping)"
+			if len(fields) >= 3 {
+				status.State = strings.ToLower(strings.Trim(fields[2], "()"))
+				break
+			}
+
+			// determine long state from the short one in case long one is not available
+			// eg "State:	S"
+			status.State = getProcLongState(fields[1][0])
+			break
```
golangcibot: redundant break statement (from `megacheck`)

processes_notwindows.go
```diff
@@ -27,90 +33,127 @@ func processes() ([]ProcStat, error) {
...
+			// extract raw long state
+			// eg "State:	S (sleeping)"
+			if len(fields) >= 3 {
+				status.State = strings.ToLower(strings.Trim(fields[2], "()"))
+				break
+			}
+
+			// determine long state from the short one in case long one is not available
+			// eg "State:	S"
+			status.State = getProcLongState(fields[1][0])
```
troian: Is it possible len(fields) == 1 or len(fields[1]) == 0? Bounds check?

requilence: The first check is enough. I use `strings.Fields()` it will not create the empty second item

PR #55, opened 2018-12-21
win: make the toasts appear longer

requilence:
Toast duration was 'short' by default, according to measurements, it is approx 5sec.
Now I set it to 'long' and it stays for 25sec

### DIFF
--- a/cmd/cagent/toast_windows.go
+++ b/cmd/cagent/toast_windows.go
@@ -30,9 +30,10 @@ func getExecutablePath() string {
 
 func sendErrorNotification(title, message string) error {
 	msg := toast.Notification{
-		AppID:   toastAppID,
-		Title:   title,
-		Message: message,
+		AppID:    toastAppID,
+		Title:    title,
+		Message:  message,
+		Duration: toast.Long,
 		Actions: []toast.Action{
 			{"protocol", "Test again", "cagent:test"},
 			{"protocol", "How to fix?", "https://kb.cloudradar.io/books/configuring-hosts/page/installing-agents#bkmrk-potential-problems-a"},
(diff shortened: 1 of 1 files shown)


requilence:
@soupdiver you are right, I renamed the the PR and set the description

soupdiver:
That's why I asked to _always_ have a description in the PR ;)

cmd/cagent/toast_windows.go
```diff
@@ -30,9 +30,10 @@ func getExecutablePath() string {
 
 func sendErrorNotification(title, message string) error {
 	msg := toast.Notification{
-		AppID:   toastAppID,
-		Title:   title,
-		Message: message,
+		AppID:    toastAppID,
+		Title:    title,
+		Message:  message,
+		Duration: toast.Long,
```
soupdiver: Add a shirt comment which effect `toast.Long` has

cmd/cagent/toast_windows.go
```diff
@@ -48,10 +49,11 @@ func sendErrorNotification(title, message string) error {
...
 func sendSuccessNotification(title, message string) error {
 	msg := toast.Notification{
-		AppID:   toastAppID,
-		Title:   title,
-		Message: message,
-		Actions: []toast.Action{},
+		AppID:    toastAppID,
+		Title:    title,
+		Message:  message,
+		Duration: toast.Long,
```
soupdiver: Add a shirt comment which effect `toast.Long` has

requilence: done

cmd/cagent/toast_windows.go
requilence: done

PR #56, opened 2018-12-22
perform lazy initialization of vmstat on fist measurements request

troian:
Lazy initialization of vmstat on on first request to get measurements.
Move `virt_.go` files to vmstat directory


pkg/monitoring/vmstat/types/types.go
```diff
@@ -0,0 +1,21 @@
+package vmstatTypes
```
golangcibot: don't use MixedCaps in package name; vmstatTypes should be vmstattypes (from `golint`)

troian: done

PR #57, opened 2018-12-27
Parse services list on linux&windows

requilence:
Available service managers:
- Systemd
- SysvInit
- Upstart
- Windows Service Manager

### DIFF
--- a/cmd/cagent/main.go
+++ b/cmd/cagent/main.go
@@ -58,7 +58,7 @@ func main() {
 
 	var serviceInstallUserPtr *string
 	var serviceInstallPtr *bool
-	
+
 	// Setup flag pointers
 	outputFilePtr := flag.String("o", "", "file to write the results (default ./results.out)")
 	cfgPathPtr := flag.String("c", cagent.DefaultCfgPath, "config file path")
--- a/cmd/cagent/main.go
+++ b/cmd/cagent/main.go
@@ -58,7 +58,7 @@ func main() {
 
 	var serviceInstallUserPtr *string
 	var serviceInstallPtr *bool
-
+	
 	// Setup flag pointers
 	outputFilePtr := flag.String("o", "", "file to write the results (default ./results.out)")
 	cfgPathPtr := flag.String("c", cagent.DefaultCfgPath, "config file path")
--- a/pkg/monitoring/services/services.go
+++ b/pkg/monitoring/services/services.go
@@ -223,7 +223,7 @@ func isUpstart() bool {
 }
 
 func setPathEnvVar(cmd *exec.Cmd) {
-	//ServiceManagerUnitd.Env = append(cmd.Env, "PATH="+os.Getenv("PATH"))
+	cmd.Env = append(cmd.Env, "PATH="+os.Getenv("PATH"))
 }
 
 func listSystemdServices() ([]map[string]string, error) {
--- a/pkg/monitoring/services/services.go
+++ b/pkg/monitoring/services/services.go
@@ -226,43 +226,41 @@ func setPathEnvVar(cmd *exec.Cmd) {
 	//ServiceManagerUnitd.Env = append(cmd.Env, "PATH="+os.Getenv("PATH"))
 }
 
-// ListServices detect the linux system manager and parse/combine results
-func ListServices() (map[string]interface{}, error) {
-	if runtime.GOOS != "linux" {
-		return nil, ErrorNotImplementedForOS
+func listSystemdServices() ([]map[string]string, error) {
+	var servicesList []map[string]string
+
+	services, err := ListSystemdServices()
+	if err != nil {
+		return []map[string]string{}, err
 	}
 
-	if isSystemd() {
-		services, err := ListSystemdServices()
-		if err != nil {
-			// in case of error lets try to query other
-			log.Errorf("[Services] Systemd appears running but failed to list a services: %s", err.Error())
-		} else {
-			var servicesList []map[string]string
-			for _, service := range services {
-				servicesList = append(servicesList,
-					map[string]string{
-						"name":         service.UnitFile,
-						"load_state":   service.LoadState,
-						"active_state": service.ActiveState,
-						"sub_state":    service.SubState,
-						"description":  service.Description,
-						"manager":      "systemd",
-					})
-			}
-			// if systemd returns results we can return here cause systemd can't work simultaneously with sysvinit/upstart
-			return map[string]interface{}{"list": servicesList}, err
-		}
+	for _, service := range services {
+		servicesList = append(servicesList,
+			map[string]string{
+				"name":         service.UnitFile,
+				"load_state":   service.LoadState,
+				"active_state": service.ActiveState,
+				"sub_state":    service.SubState,
+				"description":  service.Description,
+				"manager":      "systemd",
+			})
 	}
 
+	return servicesList, nil
+}
+
+func listSysVAndUpstartServicesCombined() []map[string]string {
 	sysVServices, err := ListSysVinitServices()
 	if err != nil {
+		// return map[string]map[string]string{}, err
 		// in case of error lets try to query other
 		log.Errorf("[Services] SysVinit: failed to list a services: %s", err.Error())
 	}
 
 	// create the map to check the services by name
 	servicesMap := map[string]map[string]string{}
+
+	// add SysV services to map
 	for _, service := range sysVServices {
 		servicesMap[service.UnitFile] = map[string]string{
 			"name":    service.UnitFile,
@@ -271,6 +269,7 @@ func ListServices() (map[string]interface{}, error) {
 		}
 	}
 
+	// if we detect Upstart also add Upstart services to map
 	if isUpstart() {
 		upstartServices, err := ListUpstartServices()
 		if err != nil {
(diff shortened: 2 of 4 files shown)


cmd/cagent/main.go
```diff
@@ -58,7 +58,7 @@ func main() {
 
 	var serviceInstallUserPtr *string
 	var serviceInstallPtr *bool
-
+
```
golangcibot: File is not `goimports`-ed (from `goimports`)

pkg/monitoring/services/services.go
```diff
@@ -0,0 +1,290 @@
...
+
+// SysVService contains the service's data parsed from service(Sysvinit) or initctl(Upstart)
+type SysVService struct {
+	UnitFile string
+	Status   string
+}
+
+// ListSystemdServices list Systemd services via systemctl
+func ListSystemdServices() ([]SystemdService, error) {
+	cmd := exec.Command("systemctl",
```
soupdiver: We already had issues with parsing the output of local binaries. Is it "safe" to do this way?

pkg/monitoring/services/services.go
```diff
@@ -0,0 +1,290 @@
...
+		}
+		if len(parts) == 0 {
+			// payload finished
+			// need to exit the read loop
+			break
+		}
+
+		rowColumnsValues := map[string]string{}
+		for colIndex, colName := range columns {
+
```
soupdiver: remove empty line

pkg/monitoring/services/services.go
```diff
@@ -0,0 +1,290 @@
...
+			if colIndex == len(columns)-1 {
+				rowColumnsValues[colName] = strings.Join(parts[colIndex:], " ")
+				break
+			}
+
+			rowColumnsValues[colName] = parts[colIndex]
+		}
+
+		services = append(services,
+			SystemdService{UnitFile: rowColumnsValues["UNIT"], LoadState: rowColumnsValues["LOAD"], ActiveState: rowColumnsValues["ACTIVE"], SubState: rowColumnsValues["SUB"], Description: rowColumnsValues["DESCRIPTION"]},
```
soupdiver: Use multi-line statement

pkg/monitoring/services/services.go
```diff
@@ -0,0 +1,290 @@
...
+		)
+	}
+
+	return services, scanner.Err()
```
soupdiver: we should check explicitly for error case and return `nil` if no error occurred.

pkg/monitoring/services/services.go
```diff
@@ -0,0 +1,290 @@
...
+		case "?":
+			state = "unknown"
+		}
+
+		services = append(services,
+			SysVService{UnitFile: parts[2], Status: state},
+		)
+	}
+
+	return services, scanner.Err()
```
soupdiver: same as before

pkg/monitoring/services/services.go
```diff
@@ -0,0 +1,290 @@
...
+		}
+
+		stateParts := strings.Split(parts[1], "/")
+
+		services = append(services,
+			SysVService{UnitFile: parts[0], Status: stateParts[1]},
+		)
+	}
+
+	return services, scanner.Err()
```
soupdiver: same as before

pkg/monitoring/services/services.go
```diff
@@ -0,0 +1,290 @@
...
+
+func isSystemd() bool {
+	if _, err := os.Stat("/run/systemd/system"); err == nil {
+		return true
+	}
+	return false
+}
+
+func isUpstart() bool {
+	if _, err := os.Stat("/sbin/upstart-udev-bridge"); err == nil {
```
soupdiver: Does the alone presence of the file _really_ indicate that upstart is being used?

pkg/monitoring/services/services.go
```diff
@@ -0,0 +1,290 @@
...
+func ListServices() (map[string]interface{}, error) {
+	if runtime.GOOS != "linux" {
+		return nil, ErrorNotImplementedForOS
+	}
+
+	if isSystemd() {
+		services, err := ListSystemdServices()
+		if err != nil {
+			// in case of error lets try to query other
+			log.Errorf("[Services] Systemd appears running but failed to list a services: %s", err.Error())
```
soupdiver: Shouldn't we return in this case?

pkg/monitoring/services/services.go
```diff
@@ -0,0 +1,290 @@
...
+	return false
+}
+
+func setPathEnvVar(cmd *exec.Cmd) {
+	// path PATH env var to be able to found right exec location
+	cmd.Env = append(cmd.Env, "PATH="+os.Getenv("PATH"))
+}
+
+// ListServices detect the linux system manager and parse/combine results
+func ListServices() (map[string]interface{}, error) {
```
soupdiver: The function is not super dramatic but I think it can be done a bit cleaner.

I suggest t to put the actual fetching for each service in its own function.
Determine which init system we're running on and then do a `switch` to call the corresponding sub-function.

Or does this not work for some reason?

pkg/monitoring/services/services.go
```diff
@@ -0,0 +1,290 @@
...
+			}
+			// if systemd returns results we can return here cause systemd can't work simultaneously with sysvinit/upstart
+			return map[string]interface{}{"list": servicesList}, err
+		}
+	}
+
+	sysVServices, err := ListSysVinitServices()
+	if err != nil {
+		// in case of error lets try to query other
+		log.Errorf("[Services] SysVinit: failed to list a services: %s", err.Error())
```
soupdiver: Do we continue in error case?

pkg/monitoring/services/services_windows.go
```diff
@@ -0,0 +1,64 @@
...
+import (
+	"context"
+	"github.com/StackExchange/wmi"
+	"time"
+)
+
+const serviceListTimeout = time.Second * 5
+
+// ErrorNotImplementedForOS exists here just for cross-platform building (cause it presented in services.go)
+var ErrorNotImplementedForOS error
```
soupdiver: Shouldn't it still contain an error message?

pkg/monitoring/services/services_windows.go
```diff
@@ -0,0 +1,64 @@
...
+}
+
+// ListServices parse Windows Service Manager
+func ListServices() (map[string]interface{}, error) {
+	ctx, cancel := context.WithTimeout(context.Background(), serviceListTimeout)
+	defer cancel()
+
+	var wmiServices []Win32_Service
+	err := wmiQueryWithContext(ctx, "Select Name,DisplayName,StartMode,State,Status from Win32_Service", &wmiServices)
+
```
soupdiver: remove empty line

pkg/monitoring/services/services.go
soupdiver: > Before we had some issues because of columns parsing, now I make it more robust because it reads the header first to match the values

👍

An idea (out of scope if this PR):
Should we think about a feature, that whenever we fail to parse some output like this, we collect and send it to our servers to we can analyse what went wrong?

pkg/monitoring/services/services.go
requilence: Not sure here. Theoretically, it may be possible that `systemd` is installed(to replace of sysVinit) and running but was not initialized properly and sysVinit still handles the services. 
I may postpone the error from here to the end of the method in case other methods is also failed

pkg/monitoring/services/services.go
requilence: Yes. SysVinit is compatible with Upstart. So we should check both

pkg/monitoring/services/services.go
requilence: Mm... Like this? 
image

The problem is that `scanner.Error()` can theoretically return an error only after some successful so we already collected some services

pkg/monitoring/services/services.go
requilence: >I suggest t to put the actual fetching for each service in its own function.

this is already done because there are `ListSystemdServices()`, `ListSysVinitServices()`, `ListUpstartServices()`
- `ListServices()`   reads, transform to map and combine(sysVinit+upstart) the different managers' services list
- systemd has different struct type (and output)
- it uses the map to combine the results(of sysVinit and upstart) by the service name

Maybe I can rework it so `ListSystemdServices()`, `ListSysVinitServices()`, `ListUpstartServices()` will return `map[string]string` instead of struct. But I don't like this because it becomes not clear when you can't fully describe what the func is excepted to return

pkg/monitoring/services/services.go
requilence: This is an error due to reading the output of intermediary-utility, so I think it may be reasonable to return both partial result and error in case it was failed.

soupdiver: ok

pkg/monitoring/services/services.go
requilence: Yes, it can be useful, but
- We will probably need some log processing setup like logstash
- I think it's better for @thorstenkramm to decide if this is OK in terms of privacy(we only send what absolutely necessary)

pkg/monitoring/services/services.go
requilence: So.. Is it OK to leave it as is if we keep partial results?

pkg/monitoring/services/services.go
soupdiver: yeah sure, I wasn't thinking about to add this here and now

pkg/monitoring/services/services.go
```diff
@@ -0,0 +1,299 @@
...
+				rowColumnsValues[colName] = strings.Join(parts[colIndex:], " ")
+				break
+			}
+
+			rowColumnsValues[colName] = parts[colIndex]
+		}
+
+		services = append(services,
+			SystemdService{
+				UnitFile: rowColumnsValues["UNIT"],
```
golangcibot: File is not `goimports`-ed (from `goimports`)

pkg/monitoring/services/services_windows.go
requilence: done

PR #58, opened 2019-01-03
decode and post hardware inventory

troian:
Decode hardware capabilities and post them alongside measurements
**Windows**: `WMI`
***nix**: `dmidecode`

On ***nix** as well adds `/etc/sudoers.d/cagent-dmidecode` to allow `cagent` execute dmidecode without prompting password


Gopkg.lock
```diff
@@ -165,6 +173,7 @@
     "github.com/shirou/gopsutil/process",
     "github.com/sirupsen/logrus",
     "github.com/stretchr/testify/assert",
+    "github.com/troian/dmidecode",
```
requilence: any reason it is not included as an internal pkg?

pkg/hwinfo/hwinfo_windows.go
```diff
@@ -0,0 +1,157 @@
...
+		return "unknown"
+	}
+}
+
+func fetchInventory() (map[string]interface{}, error) {
+	res := make(map[string]interface{})
+
+	var cpus []win32_Processor
+	query := wmi.CreateQuery(&cpus, "")
+	err := wmi.Query(query, &cpus)
```
requilence: We used the `WMIQueryWithContext` in other parts of the code. This is Windows. Nobody knows for sure how many years the WMI query could take :)
https://github.com/cloudradar-monitoring/cagent/blob/master/processes_windows.go#L23

Maybe we can move this function to `pkg/wmi/wmiutil`

pkg/hwinfo/hwinfo_windows.go
```diff
@@ -0,0 +1,157 @@
...
+	res["baseboard.manufacturer"] = baseBoard[0].Manufacturer
+	res["baseboard.serial_number"] = baseBoard[0].SerialNumber
+	if baseBoard[0].Model != nil {
+		res["baseboard.model"] = baseBoard[0].Model
+	}
+
+	var ram []win32_PhysicalMemory
+	query = wmi.CreateQuery(&ram, "")
+	if err = wmi.Query(query, &ram); err != nil {
+		return nil, fmt.Errorf("hwinfo: request ram info %s", err.Error())
```
requilence: are you sure we need to return nil here? If we were failed to retrieve RAM info we still have a lot of other info in the `res`

config.go
```diff
@@ -103,6 +105,7 @@ func NewConfig() *Config {
 		NetInterfaceExcludeRegex:        []string{},
 		NetInterfaceExcludeLoopback:     true,
 		SystemFields:                    []string{"uname", "os_kernel", "os_family", "os_arch", "cpu_model", "fqdn", "memory_total_B"},
+		HardwareInventory:               true,
```
soupdiver: Here we set to `true` but the default in the config is `false`?

pkg/hwinfo/hwinfo.go
```diff
@@ -0,0 +1,13 @@
+package hwinfo
+
+func Inventory() (map[string]interface{}, error) {
+	hw, err := fetchInventory()
+	if err != nil {
+		return nil, err
+	}
+
+	res := make(map[string]interface{})
+	res[""] = hw
```
soupdiver: index `""`?

pkg/hwinfo/hwinfo_nonwindows.go
```diff
@@ -0,0 +1,74 @@
...
+	}
+
+	dmi, err := dmidecode.Unmarshal(bufio.NewReader(&buf))
+	if err != nil {
+		return nil, fmt.Errorf("hwinfo: unmarshal dmi %s", err.Error())
+	}
+
+	res := make(map[string]interface{})
+
+	var sysArray []dmidecode.ReqBaseBoard
```
soupdiver: a slice is no Array :)

config.go
troian: typo, done

pkg/hwinfo/hwinfo_nonwindows.go

pkg/hwinfo/hwinfo_windows.go
troian: done

pkg/hwinfo/hwinfo_windows.go
troian: Ok

pkg/hwinfo/hwinfo.go
troian: Yah, looks ugly )
Reason because it resulting json output should be in form
```
hw.inventory = {
    cpu.0.Manufacture = ""
}
```

but not
```
hw.inventory.cpu.0.Manufacture = ""
```

Probably `MeasurementsMap` needs some sort of
```
func (mm MeasurementsMap) AddInnerWithPrefix(prefix string, m MeasurementsMap) MeasurementsMap {
        mm[prefix] = m
	return mm
}
```

pkg/wmi/wmi.go
```diff
@@ -53,3 +55,20 @@ func CheckOptionalFeatureStatus(feature string) (FeatureInstallState, error) {
 
 	return dst[0].InstallState, nil
 }
+
+func QueryWithContext(timeout time.Duration, query string, dst interface{}, connectServerArgs ...interface{}) error {
```
soupdiver: This function doesn't really make sense.
- if we use a `context`, we should pass it as an argument but not create the context inside the function.
- If the only usage of the `context` is the timeout, check `time.After`

pkg/hwinfo/hwinfo.go
```diff
@@ -6,8 +6,8 @@ func Inventory() (map[string]interface{}, error) {
 		return nil, err
 	}
 
-	res := make(map[string]interface{})
-	res[""] = hw
+	// res := make(map[string]interface{})
```
soupdiver: keep commented code?

troian: done

Gopkg.lock
troian: moved to github.com/cloudradar-monitoring/dmidecode as discussed in slack

soupdiver:
ping @requilence

PR #61, opened 2019-01-13
Service control with flags

troian:
start/stop/restart and check status by
cagent --service_start(stop|restart|status)
instead of running host specific commands

example

```
$ cagent --service_status
running
$ cagent --service_start
already
$ cagent --service_stop
stopped
$ cagent --service_start
started
```


cagent.go
```diff
@@ -59,14 +59,15 @@ func New(cfg *Config, version string) *Cagent {
 
 	ca.SetLogLevel(ca.Config.LogLevel)
 
+	return ca
+}
+
+func (ca *Cagent) Initialize() error {
```
soupdiver: Is there a reason why we can't have `Initialize` `New`?

cmd/cagent/main.go
```diff
@@ -105,16 +111,27 @@ func main() {
...
+	handleServiceCommand(ca, *flagServiceStatusPtr, *flagServiceStartPtr, *flagServiceStopPtr, *flagServiceRestartPtr)
+
+	handleFlagTest(*testConfigPtr, ca)
+
 	setDefaultLogFormatter()
 
+	// cagent Initialize must be called handleFlagPrintConfig and handleServiceCommand
+	// to prevent print logger setup error
+	if err = ca.Initialize(); err != nil {
+		if err != nil {
```
soupdiver: double checking for `if err != nil`

troian:
Simplifies a bit interactions for plesk extension.
We can either write lot's of bash code to detect service manager type and how to control service in terms of start/stop/status.
Instead this PR adds few commands to control cagent service over kardianos/service package which already used by cagent.
Current PR looks a bit ugly but @requilence mentioned https://github.com/jessevdk/go-flags so resulting command can be `cagent service start`

soupdiver:
> Simplifies a bit interactions for plesk extension.
> We can either write lot's of bash code to detect service manager type and how to control service in terms of start/stop/status.
> Instead this PR adds few commands to control cagent service over kardianos/service package which already used by cagent.
> Current PR looks a bit ugly but @requilence mentioned https://github.com/jessevdk/go-flags so resulting command can be `cagent service start`

ok, I see.

cagent.go

cmd/cagent/main.go
```diff
@@ -105,7 +111,15 @@ func main() {
 
 	handleFlagPrintConfig(*printConfigPtr, cfg)
 
-	setDefaultLogFormatter()
+	if ((serviceInstallPtr == nil) || ((serviceInstallPtr != nil) && (!*serviceInstallPtr))) &&
```
troian: looks like https://github.com/jessevdk/go-flags suggests itself as this now look too ugly.
@soupdiver @requilence ?

requilence: agree. Now the CLI API becomes to look ugly 
image

soupdiver: I agree, it get's messy.
An opinion on grouping? E.g. `cagent service start` and maybe the other options could also be grouped... do we need backwards compatibility in this case?

cmd/cagent/main.go
```diff
@@ -526,11 +604,18 @@ func getServiceFromFlags(ca *cagent.Cagent, configPath, userName string) (servic
...
 
-func setDefaultLogFormatter() {
+func setDefaultLogFormatter(ca *cagent.Cagent) {
 	tfmt := log.TextFormatter{FullTimestamp: true}
 	if runtime.GOOS == "windows" {
 		tfmt.DisableColors = true
 	}
 
 	log.SetFormatter(&tfmt)
+
```
requilence: the name of the function now doesn't show what is happening inside

troian: @requilence It still configures default log formatter. Isn't ?

requilence: from the other side, the function with name `setDefaultLogFormatter` now adds the file log hook which seems not quite expected

soupdiver: > now adds the file log hook which seems not quite expected

agree

troian: done

soupdiver:
@requilence Can we get this merged?
Nearly three weeks old

PR #62, opened 2019-01-14
Docker integration

requilence:
- List running docker containers
- Append container name to the procs.list


pkg/monitoring/docker/docker.go
```diff
@@ -0,0 +1,65 @@
+package docker
+
+import (
+	"strings"
+
+	"github.com/fsouza/go-dockerclient"
```
golangcibot: File is not `goimports`-ed (from `goimports`)

processes.go
```diff
@@ -6,12 +6,14 @@ import (
 	log "github.com/sirupsen/logrus"
 )
 
+//todo: make all JSON keys lowercase
```
soupdiver: A reason why this is todo?

processes_notwindows.go
```diff
@@ -107,6 +112,23 @@ func processesFromProc() ([]ProcStat, error) {
 			stat.Cmdline = strings.Replace(string(bytes.TrimRight(cmdline, "\x00")), "\x00", " ", -1)
 		}
 
+		cgroupFilepath := getHostProc() + "/" + pidString + "/cgroup"
+		cgroup, err := readProcFile(cgroupFilepath)
+		if err != nil && err != errorProcessTerminated {
+			log.Errorf("[PROC] failed to read cgroup(%s): %s", cgroupFilepath, err.Error())
```
soupdiver: can we call `continue` here and then un-indent  the `else` branch?

processes.go
requilence: I want to include this in the separate PR because to provide context
This is partly backward-incompatible change and will require some  changes from API side

Gopkg.toml
```diff
@@ -60,3 +60,15 @@
 [[constraint]]
   name = "github.com/stretchr/testify"
   version = "1.2.2"
+
+[[override]]
+  name = "github.com/docker/docker"
```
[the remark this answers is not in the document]
requilence: In fact, it doesn't require root docker/docker. Only some of sub pkgs: https://github.com/cloudradar-monitoring/cagent/blob/6dcb003285c33e0139d76ebd4d865734024e9cad/Gopkg.lock#L52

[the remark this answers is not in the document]
soupdiver: Will still all the `github.com/docker/docker` repo end up in our `vendor` or just the things we actually use?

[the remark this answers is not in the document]
requilence: Nope, only the part of it mentioned in the Gopkg.lock section above

processes_notwindows.go
requilence: no, we can't, it still needs to add proc stat to slice later
https://github.com/cloudradar-monitoring/cagent/blob/6dcb003285c33e0139d76ebd4d865734024e9cad/processes_notwindows.go#L132

pkg/monitoring/docker/docker.go
requilence: I have no difference after gofmt+goimports
You are badass @golangcibot 😝

processes_notwindows.go
soupdiver: ok, but we can still remove the `else` by inverting the logic and having `procs = append(procs, stat)` in before the continue as well right?
I wanna minimise the `if/else` occurrence in the code

pkg/monitoring/docker/docker.go
soupdiver: haha... just tried myself and for me there is nothing in this file as well 🤷‍♀️

processes_notwindows.go
requilence: we can copy append, yes. But later we may add something after it and introduce a bug

https://github.com/cloudradar-monitoring/cagent/blob/f366c916f3dcbd0461dd433581cab698d1181297/processes_notwindows.go#L99-L132

requilence: we can convert inner-range code into separate func and save one level :)
We can also convert each proc file processing into separate func. But for other funcs this will gives overhead imho

soupdiver: ah okay... well then let's keep it as is

PR #63, opened 2019-01-16
doc: mention plesk depends on cagent config path

troian:
Plesk extension management (enable|disable) relies on default config path
Warn plesk must be update prior to do it in the cagent

### DIFF
--- a/pkg-scripts/postinstall.sh
+++ b/pkg-scripts/postinstall.sh
@@ -1,3 +1,7 @@
 #!/bin/sh
+
+# WARN: path to config is referenced by plesk extension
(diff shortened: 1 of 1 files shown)


pkg-scripts/postinstall.sh
```diff
@@ -1,3 +1,7 @@
 #!/bin/sh
+
+# WARN: path to config is referenced by plesk extension
+#       when enabling/disabling plugin thus plesk extension
```
soupdiver: `thus` -> `this`

I don't understand the warning.

troian: PR is not necessary anymore

PR #65, opened 2019-01-24
fix: do not fail if dmidecode is not present

troian:
Report `cagent.success: 1` if `dmidecode` tool is not present if the system

### DIFF
--- a/handler.go
+++ b/handler.go
@@ -147,6 +147,7 @@ func (ca *Cagent) PostResultsToHub(result Result) error {
 
 func (ca *Cagent) GetAllMeasurements() (MeasurementsMap, error) {
 	var errs []string
+
 	var measurements = make(MeasurementsMap)
 
 	cpum, err := ca.CPUWatcher().Results()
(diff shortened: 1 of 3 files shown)


handler.go
```diff
@@ -147,6 +147,7 @@ func (ca *Cagent) PostResultsToHub(result Result) error {
 
 func (ca *Cagent) GetAllMeasurements() (MeasurementsMap, error) {
 	var errs []string
+
```

pkg/hwinfo/hwinfo_nonwindows.go
```diff
@@ -11,7 +11,20 @@ import (
 	"github.com/cloudradar-monitoring/dmidecode"
 )
 
+func isCommandAvailable(name string) bool {
+	cmd := exec.Command("/bin/sh", "-c", "command -v "+name)
```
soupdiver: Does this try to execute the provided command?
Can this be a security risk?

handler.go

pkg/hwinfo/hwinfo_nonwindows.go
troian: this is builtin shell command

PR #68, opened 2019-01-25
Disable docker on windows

requilence:
- Currently, it tries to connect to the Unix socket on windows, producing errors
- It is still possible to use it on windows via a named pipe, but @thorstenkramm said that we should disable docker on Windows as this is is a pretty rare case.
- As advantage, this saves some binary size on Windows :)

### DIFF
--- a/pkg/monitoring/docker/docker_windows.go
+++ b/pkg/monitoring/docker/docker_windows.go
@@ -7,5 +7,5 @@ func ListContainers() (map[string]interface{}, error) {
 }
 
 func ContainerNameByID(_ string) (string, error) {
-	return "", nil
+	return "", ErrorNotImplementedForOS
 }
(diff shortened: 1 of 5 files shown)


pkg/monitoring/docker/docker_windows.go
```diff
@@ -0,0 +1,11 @@
+// +build windows
+
+package docker
+
+func ListContainers() (map[string]interface{}, error) {
+	return nil, ErrorNotImplementedForOS
+}
+
+func ContainerNameByID(_ string) (string, error) {
```
soupdiver: Shouldn't this return `ErrorNotImplementedForOS ` as `err` as well?

requilence: ContainerNameByID() only called by `processes_notwindows.go`
But I fixed it to make it looks better :)

PR #69, opened 2019-01-27
CPU threshold notifier & top integration

requilence:
- monitor CPU util metrics using new ThresholdNotifier at CPU Watcher
- Start/Stop Top measurement when it exceed
- Sort, cut and provide top results

PR #71, opened 2019-01-28
Fix top

### LINKED ISSUE #70: top pkg issues
I have some problems with top pkg that needs to be solved for https://github.com/cloudradar-monitoring/cagent/pull/69 to be merged

1. Need to add lowercase-naming json struct tags for `ProcessInfo` struct. Any reason you renamed process `Name` into `Identifier`? This could probably confuse both devs & users.

2. I got a lot of logs like this on Linux. I didn't dig  why this procs not exist (maybe very short-lived) but probably need to omit them
image

3. I made a sorting of `GetProcesses()` on my side, but it feels like it should be done on `top` side. The same way it is done in the `top` utility by default.

soupdiver:
Fixes #70 

Should be merged before #69

### DIFF
--- a/config.go
+++ b/config.go
@@ -175,6 +175,8 @@ func TryUpdateConfigFromFile(cfg *Config, configFilePath string) error {
 		return err
 	}
 
+	// log.Printf("WARP: %+v", cfg)
+
 	return nil
 }
 
@@ -265,13 +267,11 @@ func HandleAllConfigSetup(configFilePath string) (*Config, error) {
 		if strings.Contains(err.Error(), "cannot load TOML value of type int64 into a Go float") {
 			return nil, fmt.Errorf("Config load error: please use numbers with a decimal point for numerical values")
 		}
-
 		return nil, fmt.Errorf("Config load error: %s", err.Error())
 	}
 
 	if err = cfg.validate(); err != nil {
 		return nil, err
 	}
-
 	return cfg, nil
 }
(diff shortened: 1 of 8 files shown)


handler.go
```diff
@@ -310,7 +310,22 @@ func (ca *Cagent) RunOnce(outputFile *os.File) error {
 	return ca.ReportMeasurements(measurements, outputFile)
 }
 
-func (ca *Cagent) Run(outputFile *os.File, interrupt chan struct{}) {
+func (ca *Cagent) Run(outputFile *os.File, interrupt chan struct{}, cfg *Config) {
+	// Start process utilisation monitoringi in the background
```
golangcibot: `utilisation` is a misspelling of `utilization` (from `misspell`)

config.go
```diff
@@ -66,7 +66,16 @@ type Config struct {
 
 	VirtualMachinesStat []string `toml:"virtual_machines_stat" comment:"default ['hyper-v'], available options 'hyper-v'"`
 
-	HardwareInventory bool `toml:"hardware_inventory" comment:"default true"`
+	HardwareInventory      bool                   `toml:"hardware_inventory" comment:"default true"`
+	CPUUtilisationAnalysis CPUUtilisationAnalysis `toml:"cpu_utilisation_analysis"`
+}
+
+type CPUUtilisationAnalysis struct {
```
requilence: Did you intentionally put here the struct from my PR?

soupdiver: The decision was made later to move the detection logic to `cpuWatcher`. I had already added this feature to `Top`.
Wasn't sure about the other PR and therefor left this here for now.
I guess I can remove this struct then?

requilence: Ah, got it. Yes, I think so.. I thought that it can be useful to have the analyzer as a separate pkg. So it could be possible attach more pkg like `top` there later

requilence:
Should also remove from `config_test.go` or lets merge and then fix tests after #69?

soupdiver:
addressed issues @requilence

PR #73, opened 2019-01-28
Improve services list

requilence:
To fulfill DEV-585 discussion.
- detect auto_start(*see notes below)
- show only autostarting service by default(discover_autostarting_services_only in config)
- use `state` field for all of system managers to make filtering in the UI easier
- query `delayed autostart` status on win
- make win values lowercase 

*auto_start detection
- On Windows they have the `run=auto`
- On Linux SystemD I'm filtering only __STATE=enabled__ services from `systemctl --type=service --all --no-pager --plain --output=json list-unit-files`
- TODO: SysVinit/upstart currently assuming all services as auto-start. Will need to solve this by parsing unit files or list&correlate symlinks from `/etc/rc*.d`

### DIFF
--- a/config.go
+++ b/config.go
@@ -67,6 +67,8 @@ type Config struct {
 	VirtualMachinesStat []string `toml:"virtual_machines_stat" comment:"default ['hyper-v'], available options 'hyper-v'"`
 
 	HardwareInventory bool `toml:"hardware_inventory" comment:"default true"`
+
+	DiscoverAutostartingServicesOnly bool `toml:"discover_autostarting_services_only" comment:"default true"`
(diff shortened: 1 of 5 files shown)


soupdiver:
What about the todo?
`TODO: SysVinit/upstart currently assuming all services as auto-start. Will need to solve this by parsing unit files or list&correlate symlinks from /etc/rc*.d`

requilence:
@soupdiver we talked about this with Thorsten and decided to go as is with for the next release(it could include non-autostart services for Upstart&SysV-managed systems) and then improve it using runlevels symlnks

PR #74, opened 2019-01-28
ignore failed processes, if we log the errors we get way too many errors

soupdiver:
Fix for #71 

Ignore the processes completely in case something goes wrong. Likely for short-lived processes

### DIFF
--- a/pkg/monitoring/top/top_nonwindows.go
+++ b/pkg/monitoring/top/top_nonwindows.go
@@ -3,6 +3,7 @@
 package top
 
 import (
+	"sync"
 	"sync/atomic"
 	"time"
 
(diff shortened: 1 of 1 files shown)


pkg/monitoring/top/top_nonwindows.go
```diff
@@ -23,21 +21,25 @@ func (t *Top) GetProcesses() ([]*ProcessInfo, error) {
...
 		// Run in background because the call to Percent blocks for the duration
 		go func(p *process.Process) {
 			load, err := p.Percent(time.Second * 1)
-			if err != nil && !strings.Contains(err.Error(), "no such file") {
-				// Ignore the file not not found case to avoid a lot of error messages for
-				// short lived processes
-				log.Printf("Err getting Percent: %s", err)
+			if err != nil {
+				// If we log the error in this place, we get _a lot of_ messages
+				skipped++
```
requilence: it's not a good idea to increment the counter inside goroutine inside goroutine. It is not safe.

It's better to use sync.WaitGroup and close the channel after

soupdiver: > It is not safe

Incrementing is an atomic operation, there is no `if` with a following instruction or similar. Therefor it's ok to do this in a goroutine

requilence: Unfortunately incrementing is not an atomic operation. That's why sync.atomic exists

soupdiver: Addressed

soupdiver:
> please, can you rework to make it use WaitGroup instead of counter

`WaitGroup` is to wait for background tasks to complete. In our case we need the results of these operations as well and the idiomatic way to synchronise these things in go are channels, which is what I use. When switching to a waitGroup, then we also have to use a mutex to organise processing of the results. 
So I don't see the advantage of a `waitGroup` in this case 🤔

PR #78, opened 2019-01-31
Deb/rpm and -t fixes

requilence:
- Fix typo (succeed->succeeded)
- Show instruction on how to restart the service only in case there is an error and service is actually installed
- Improve the pkg installer output

### DIFF
--- a/cmd/cagent/main.go
+++ b/cmd/cagent/main.go
@@ -437,10 +437,10 @@ func handleFlagTest(testConfig bool, ca *cagent.Cagent) {
 	}
 
 	if runtime.GOOS == "windows" {
-		_ = sendSuccessNotification("Cagent connection test succeed", "")
+		_ = sendSuccessNotification("Cagent connection test succeeded", "")
 	}
 
-	fmt.Printf("HUB connection test succeed and credentials are correct!\n")
+	fmt.Printf("HUB connection test succeeded and credentials are correct!\n")
 	os.Exit(0)
 }
 
--- a/cmd/cagent/main.go
+++ b/cmd/cagent/main.go
@@ -379,10 +379,9 @@ func handleFlagServiceInstall(ca *cagent.Cagent, systemManager service.System, s
 
 	if ca.Config.HubURL == "" {
 		fmt.Printf(`*** Attention: 'hub_url' config param is empty.\n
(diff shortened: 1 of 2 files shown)


cmd/cagent/main.go
```diff
@@ -433,14 +426,32 @@ func handleFlagTest(testConfig bool, ca *cagent.Cagent) {
...
+
+		status, err := systemService.Status()
+		if err != nil {
+			// service seems not installed
+			// no need to show the tip on how to restart it
+			os.Exit(1)
+		}
+
+		systemManager := service.ChosenSystem()
+		if status == service.StatusRunning || status == service.StatusStopped {
```
soupdiver: Is this the check for an incorrect config?
Is `if status == service.StatusRunning` correct then? We check if the service is running, if yes we tell the user to fix the config and restart? Does the service start with an invalid config?

handler.go
```diff
@@ -56,13 +56,13 @@ func (ca *Cagent) initHubHTTPClient() {
...
 	if ca.Config.HubURL == "" {
-		return fmt.Errorf("please set the hub_url config param")
+		return fmt.Errorf("please fill config with hub_url, hub_user and hub_password from your Cloudradar account")
 	}
 
 	var u *url.URL
 	var err error
 	if u, err = url.Parse(ca.Config.HubURL); err != nil {
-		return fmt.Errorf("can't parse hub_url: %s", err.Error())
+		return fmt.Errorf("can't parse hub_url: %s. Make sure to put the right params from your Cloudradar account", err.Error())
```
soupdiver: `params` => `url` maybe?

requilence: if we got the error parsing hub_url it doesn't mean that user put the right hub_user and hub_password. So that's why I have mention `params`

cmd/cagent/main.go
soupdiver: > it can be running(e.g. in case hub_user/hub_password is invalid or hub_url is reachable)

handler.go
soupdiver: yes, but if `url.Parse(ca.Config.HubURL)` fails we know _exactly_ that the url is wrong right? So we can give a specific error and not just a generic one or is there something later in the flow I miss right now?

requilence: the error starts with: `can't parse hub_url`

soupdiver: ok, fine...

PR #79, opened 2019-01-31
Auto-create config dir if not exists

requilence:
For example It was a fatal error when trying to run the cagent on linux system without `/etc/cagent` directory presented


config.go
```diff
@@ -212,9 +212,17 @@ func GenerateDefaultConfigFile(mvc *MinValuableConfig, configFilePath string) er
...
+		err := os.MkdirAll(configPathDir, os.ModePerm)
+		if err != nil {
+			return fmt.Errorf("failed to auto-create the default Config file directory '%s': %s", configPathDir, err.Error())
+		}
+	}
+
 	var f *os.File
 	if f, err = os.OpenFile(configFilePath, os.O_WRONLY|os.O_CREATE, 0666); err != nil {
-		return fmt.Errorf("failed to create the default Config file: '%s'", configFilePath)
+		return fmt.Errorf("failed to create the default Config file at '%s': %s", configFilePath, err.Error())
```
soupdiver: just a small note, why is everything non-capital _but_ `Config` was capitalised?

requilence: it was made this way by @troian. But yeah, its better to make it lowercase

soupdiver: Are we going to update this?

requilence: fixed

PR #80, opened 2019-01-31
improve docker missing error handling

requilence:
The check I have before was not enough, because for some cases docker lib returns the different error.
So the new plan is:
- check if executable `docker` is presented at `$PATH`. 
- If `docker` is not presented on system AND docker request wasn't successful within the session we can omit the error

### DIFF
--- a/cagent.go
+++ b/cagent.go
@@ -28,7 +28,7 @@ type Cagent struct {
 	fsWatcher            *FSWatcher
 	netWatcher           *NetWatcher
 	windowsUpdateWatcher *WindowsUpdateWatcher // nolint: structcheck,megacheck
-	docker               *docker.DockerWatcher
+	dockerWatcher        *docker.Watcher
 
 	vmstatLazyInit sync.Once
 	vmWatchers     map[string]vmstattypes.Provider
@@ -45,7 +45,7 @@ func New(cfg *Config, cfgPath string, version string) *Cagent {
 		ConfigLocation: cfgPath,
 		version:        version,
 		vmWatchers:     make(map[string]vmstattypes.Provider),
-		docker:         &docker.DockerWatcher{},
+		dockerWatcher:  &docker.Watcher{},
 	}
 
 	if rootCertsPath != "" {
--- a/cagent.go
+++ b/cagent.go
@@ -45,7 +45,14 @@ func New(cfg *Config, cfgPath string, version string) *Cagent {
 		ConfigLocation: cfgPath,
 		version:        version,
 		vmWatchers:     make(map[string]vmstattypes.Provider),
-		dockerWatcher:  &docker.Watcher{},
+	}
+
+	if dw, err := docker.New(); err != nil {
+		if err != docker.ErrorNotImplementedForOS {
+			log.Errorf("[docker] init error: %s", err)
+		}
+	} else {
+		ca.dockerWatcher = dw
 	}
 
 	if rootCertsPath != "" {
--- a/cagent.go
+++ b/cagent.go
@@ -45,6 +45,7 @@ func New(cfg *Config, cfgPath string, version string) *Cagent {
 		ConfigLocation: cfgPath,
 		version:        version,
 		vmWatchers:     make(map[string]vmstattypes.Provider),
+		docker:         &docker.DockerWatcher{},
 	}
 
 	if rootCertsPath != "" {
--- a/cagent.go
+++ b/cagent.go
@@ -10,6 +10,7 @@ import (
 	"strings"
 	"sync"
 
+	"github.com/cloudradar-monitoring/cagent/pkg/monitoring/docker"
 	"github.com/cloudradar-monitoring/cagent/pkg/monitoring/vmstat"
 	vmstattypes "github.com/cloudradar-monitoring/cagent/pkg/monitoring/vmstat/types"
 
@@ -27,9 +28,11 @@ type Cagent struct {
 	fsWatcher            *FSWatcher
 	netWatcher           *NetWatcher
 	windowsUpdateWatcher *WindowsUpdateWatcher // nolint: structcheck,megacheck
-	vmstatLazyInit       sync.Once
-	vmWatchers           map[string]vmstattypes.Provider
-	hwInventory          sync.Once
+	docker               *docker.DockerWatcher
+
+	vmstatLazyInit sync.Once
+	vmWatchers     map[string]vmstattypes.Provider
+	hwInventory    sync.Once
 
 	rootCAs *x509.CertPool
 
--- a/handler.go
+++ b/handler.go
@@ -17,7 +17,6 @@ import (
 	log "github.com/sirupsen/logrus"
 
 	"github.com/cloudradar-monitoring/cagent/pkg/hwinfo"
-	"github.com/cloudradar-monitoring/cagent/pkg/monitoring/docker"
 	"github.com/cloudradar-monitoring/cagent/pkg/monitoring/services"
 	"github.com/cloudradar-monitoring/cagent/pkg/monitoring/vmstat"
 )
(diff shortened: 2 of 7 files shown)


pkg/monitoring/docker/docker.go
```diff
@@ -3,6 +3,8 @@
 package docker
 
 import (
+	"fmt"
```
golangcibot: File is not `goimports`-ed (from `goimports`)

pkg/monitoring/docker/docker.go
```diff
@@ -22,10 +23,17 @@ func ListContainers() (map[string]interface{}, error) {
...
 	}
+	var dockerExecExists bool
+	if _, err := exec.LookPath("docker"); err == nil {
+		dockerExecExists = true
+	}
 
 	containers, err := client.ListContainers(docker.ListContainersOptions{All: false})
-	if err != nil && err == docker.ErrConnectionRefused && !connectionSucceedOnce {
-		// do not produce error if Docker is missing and wasn't successfully connected in this session before
+	if err != nil &&
```
soupdiver: Can we simplify this statement? Maybe split into two `if`?
Is the check against `connectionSucceedOnce` necessary? It's only declared but never accessed until this point

requilence:
I made more tests on some machines:
- with docker installed but not running
- without docker at all
- with docker stopped at the middle of cagent execution

Every case seems to handle properly now

pkg/monitoring/docker/docker.go
requilence: connectionSucceedOnce is set to true in case of successful docker query

soupdiver: > connectionSucceedOnce is set to true in case of successful docker query

But only at line 44 which is _after_ this check.

requilence: Yes, I know how it works and tried to do this way at first but
- I needed to store this struct at the Cagent struct
- I needed to rework `processesFromProc()` and `processes()` to pass the Cagent struct in order to run `ContainerNameByID`. I think it is not clear to use methods when it is possible to solve the task with just a func

soupdiver: > With the post-inspection finding nothing to cause the engines to fail, engineers are saying the likely cause for the shutdown is a software issue onboard the aircraft.

What exactly is not clear?

soupdiver: > And in this case, the price is too high

Which price exactly?

requilence: I've just listed the reasons :)
`processes()` and `processesFromProc()` are clear and independent as they don't need to have Cagent instance passed in order to provide results.

DO you think it is ok to rewrite them all to become methods?

soupdiver: > DO you think it is ok to rewrite them all to become methods?

yes. That's why I did the `Top` package how I did it. No global state, a central struct, `Top` in this case, which can hold state and other private members but still nicely groups the whole functionality together. If you're looking for anything related it's all within that one package and not together with 10+ other files in one package without a struct that holds things together.

In the end the same applies for `Config`, `CPUWatcher` etc.

requilence: take a look. I've just reworked it

cagent.go
```diff
@@ -27,9 +28,11 @@ type Cagent struct {
 	fsWatcher            *FSWatcher
 	netWatcher           *NetWatcher
 	windowsUpdateWatcher *WindowsUpdateWatcher // nolint: structcheck,megacheck
-	vmstatLazyInit       sync.Once
-	vmWatchers           map[string]vmstattypes.Provider
-	hwInventory          sync.Once
+	docker               *docker.DockerWatcher
```
soupdiver: name it `dockerWatcher`

handler.go
```diff
@@ -256,7 +256,10 @@ func (ca *Cagent) GetAllMeasurements() (MeasurementsMap, error) {
 
 	measurements = measurements.AddWithPrefix("services.", servicesList)
 
-	containersList, err := docker.ListContainers()
+	if ca.docker == nil {
```
soupdiver: shouldn't we do that on `cagent` creation?

pkg/monitoring/docker/docker.go
```diff
@@ -11,29 +12,47 @@ import (
...
 		}
 	}
+	return nil
+}
 
-	containers, err := client.ListContainers(docker.ListContainersOptions{All: false})
-	if err != nil && err == docker.ErrConnectionRefused && !connectionSucceedOnce {
-		// do not produce error if Docker is missing and wasn't successfully connected in this session before
+func (dw *DockerWatcher) ListContainers() (map[string]interface{}, error) {
+	err := dw.initClient()
```
soupdiver: Is it for performance reason that we create the client so late? Otherwise we should init the client on creation of `DockerWatcher`

pkg/monitoring/docker/docker.go
```diff
@@ -11,29 +12,47 @@ import (
...
+	}
+
+	var dockerExecExists bool
+	if _, err := exec.LookPath("docker"); err == nil {
+		dockerExecExists = true
+	}
+
+	containers, err := dw.client.ListContainers(docker.ListContainersOptions{All: false})
+	if err != nil &&
+		!dockerExecExists && // docker executable not exists in the system
```
soupdiver: Extraction of `DockerWatcher` 👍 

Can we also simplify the condition here? Maybe split into two `if`?

pkg/monitoring/docker/docker.go
requilence: the reason is that we don't have `New` method. If I put the client init only inside the `New()` it could be unclear why you can do `docker.New()` without params, but can't just use struct
```
d := docker.New()
d.ListContainers()
```

but can't do 
```
d := &DockerWatcher{}
d.ListContainers()
```

handler.go
requilence: done

pkg/monitoring/docker/docker.go
```diff
@@ -11,29 +12,47 @@ import (
 
 const dockerEndpointAddress = "unix:///var/run/docker.sock"
 
-var client *docker.Client
-var connectionSucceedOnce bool
+type DockerWatcher struct {
```
golangcibot: type name will be used as docker.DockerWatcher by other packages, and that stutters; consider calling this Watcher (from `golint`)

pkg/monitoring/docker/docker.go
soupdiver: > could be unclear why you can do `docker.New()` without params, but can't just use struct

_Exactly_ this is the reason. You should not use the struct directly except you do all the initialisation manually as well.
A comment in the code pointing this out should be fine... but I think it's sort of the norm that you have a `New` function in case the struct needs some init code

requilence: done

pkg/monitoring/docker/docker.go
requilence: I have added the comments. Let's leave it as is. I didn't find the way to split it elegantly

soupdiver: At least clean up the `else if` then. This is a pattern I already mentioned many times.
If you have a `return` inside the `if` branch you _don't_ have to put an `else if` afterwards. Omit the `else` because if the first branch is triggered, the function/method returns anyway and therefor the `else` is not needed.

Please try to keep that in mind for the future.

requilence: Right, `else` is not necessary here. But I think it improves the readability.
If `else` is omitted when someone the code fluently he can think that the second error handling is the only one

soupdiver: > But I think it improves the readability in this case.

I disagree. As you said, it's not needed and therefor should be removed.
The first `if` checks for the specific error and the second `if` checks for the general error.
Add a comment to each `if` to describe what this check does, done.

> he can think that the second error handling is the only one

So you write the code for someone not reading the code properly?

requilence: take a look, I fixed it by moving it into nested if (but I know you hate them 😅)

soupdiver: ok... whatever, don't take my advice

PR #81, opened 2019-01-31
Windows WMI fix: make Description *string

requilence:
WMI returns nil Description for some services

### DIFF
--- a/pkg/monitoring/services/services_windows.go
+++ b/pkg/monitoring/services/services_windows.go
@@ -18,7 +18,7 @@ var ErrorNotImplementedForOS error
 type Win32_Service struct {
 	Name             string
 	DisplayName      string
-	Description      string
+	Description      *string
 	StartMode        string
 	State            string
 	Status           string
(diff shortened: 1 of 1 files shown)


pkg/monitoring/services/services_windows.go
```diff
@@ -18,7 +18,7 @@ var ErrorNotImplementedForOS error
 type Win32_Service struct {
 	Name             string
 	DisplayName      string
-	Description      string
+	Description      *string
```
soupdiver: Why do we have to make it a pointer?

pkg/monitoring/services/services_windows.go
```diff
@@ -46,7 +46,7 @@ func ListServices(autostartOnly bool) (map[string]interface{}, error) {
 	defer cancel()
 
 	var wmiServices []Win32_Service
-	err := wmiQueryWithContext(ctx, "Select Name,DisplayName,Description,StartMode,State,State,DelayedAutoStart from Win32_Service", &wmiServices)
+	err := wmiQueryWithContext(ctx, "Select Name,DisplayName,Description,StartMode,State,Status,DelayedAutoStart from Win32_Service", &wmiServices)
```
soupdiver: Why change State => Status?

pkg/monitoring/services/services_windows.go
requilence: because of this
image

pkg/monitoring/services/services_windows.go
requilence: it was a typo

Issue #76, opened 2019-01-29
Top pkg questions


requilence:
@soupdiver I have some questions on the Linux top calculation approach
Currently, it works like this:
1) get the list of process (listing `/proc/` dir)
2) for each process, you query the CPU times
3) sleep 1 second
4) get the new CPU times
5) calculate the delta and percent
6) wait 1 second
7) repeat from 1)

Problems: 
- it queries procfs actively every 1 second. I didn't make detailed profiling, but it seems like adding additional 1-2% of CPU load on my system. And it doesn't seem to add any precision for calculations
- You can lose some spikes inside the 6) it's better to use non-blocking mode (https://github.com/shirou/gopsutil/blob/master/process/process.go#L170)

soupdiver:
> it queries procfs actively every 1 second. I didn't make detailed profiling, but it seems like adding additional 1-2% of CPU load on my system

Sounds likely... calculating the process stats does consume CPU power. When running `htop` on my machine it takes around 0.5%.
 Not sure if this on itself is a problem, since the detailed monitoring is just active for some (short) amount of time.

> And it doesn't seem to add any precision for calculations

Have you done some tests to prove that claim or why does it seem to add no precision?

> You can lose some spikes inside the 6)

Why?

> it's better to use non-blocking mode 

Why is it better to use the non-blocking call? Then we have to do the `time.Sleep` inside our code. I'm not sure if this will change something? 🤔

requilence:
> Have you done some tests to prove that claim or why does it seem to add no precision?

Under the hood, it reads CPU time from /proc/<id>/stat (measured in clock ticks), not the percentage. So if you read once in a minute you can get the same(or even more precise because you don't lose on rounding) results as reading the CPU times every second, calculating Percentage and found the average.

> Why?

It loose spikes because when you wait 1 second some process can consume the CPU heavily

> Why is it better to use the non-blocking call? Then we have to do the time.Sleep inside our code. I'm not sure if this will change something? 🤔

Because now you are doing both. Sleeping inside the gopsutil(https://github.com/shirou/gopsutil/blob/master/process/process.go#L170) and sleeping outside of it.

requilence:
top/htop reads CPU times every 2 second to update the percentage in the UI frequently. In our case by default we need to provide results once in 120 seconds

soupdiver:
> Because now you are doing both.

Agree, that might be a dev leftover 🤔 
One sleep should be enough .

[...]

PR #83, opened 2019-01-31
Top fixes

soupdiver:
See #76

### DIFF
--- a/pkg/monitoring/top/top.go
+++ b/pkg/monitoring/top/top.go
@@ -204,7 +204,7 @@ func (t *Top) HighestLoad() (string, float64, float64, float64, float64) {
 			avg5 = Avg5(pi)
 		}()
 		go func() {
-			avg1 = Avg15(pi)
+			avg15 = Avg15(pi)
 		}()
 		wg.Wait()
 	}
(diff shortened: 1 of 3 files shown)


requilence:
What about other things we talked
- Because of sorting, using the `Load` we will not likely see the short-lived high-spiked process
- Querying every second is too frequently 

I propose: 
- Add `Load1`(avg1) metric and use it for sorting
- Query every 5 seconds instead of 1 second. Because even some proc that lived 5 sec made a 100% CPU load it will result in `avg1=8%` so the user will likely ignore this

soupdiver:
> What about other things we talked

Not mentioned in the issue

soupdiver:
> Add Load1(avg1) metric and use it for sorting

done

soupdiver:
> Query every 5 seconds instead of 1 second.

Done

pkg/monitoring/top/top.go
```diff
@@ -91,6 +94,7 @@ func (t *Top) startMeasureProcessLoad(interval time.Duration) {
 					PID:        p.PID,
 					Command:    p.Command,
 					Identifier: p.Name,
+					Load1:      ring.New(60),
```
requilence: you should divide ring.New arg with `interval`

soupdiver: that is correct 👍 
done

requilence:
@soupdiver maybe it is a good idea to round float with 2 decimal place? 
```
"cpu_analyser.top":[{"name":"stress","pid":19685,"command":"stress --cpu 2 --timeout 600","load":32.00714496243167,"load1":27.151941315673664,"load5":27.718735561029046,"load15":27.76807674163746
```

PR #85, opened 2019-02-07
rework docker

requilence:
Problem: 
We have realized that docker's UNIX socket is not readable for non-privileged `cagent` user we used for system service on Linux.

Solution:
Use existing`docker` executable instead of UNIX socket connection via API and add `cagent` to `sudoers.d` to be able to execute `docker` commands with sudo

This PR also fixes the broken go dep files


handler.go
```diff
@@ -14,6 +14,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/cloudradar-monitoring/cagent/pkg/monitoring/docker"
```

pkg/monitoring/docker/docker.go
```diff
@@ -3,61 +3,107 @@
...
+func isDockerAvailable() bool {
+	if _, err := exec.LookPath("docker"); err != nil {
+		return false
 	}
 
-	return &Watcher{client: client}, nil
+	return true
+}
+
+type DockerPSOutput struct {
```
golangcibot: type name will be used as docker.DockerPSOutput by other packages, and that stutters; consider calling this PSOutput (from `golint`)

PR #86, opened 2019-02-07
improve win UI

requilence:
see DEV-682 comments
- make the separate method `TestHubWinUI()` for testing within windows UI. It returns more clear errors and doesn't mention config file
- fix starting of stopped service
- rename windows shortcut in start menu

### DIFF
--- a/cmd/cagent/winui.go
+++ b/cmd/cagent/winui.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"strings"
 	"time"
 
 	"github.com/lxn/walk"
@@ -88,13 +89,18 @@ func (ui *UI) TestSaveReload(testOnly bool) {
 	saveButtonText := ui.SaveButton.Text()
 	ui.SaveButton.SetEnabled(false)
 	ui.SaveButton.SetText("Testing...")
-	err := ui.ca.TestHub()
+	err := ui.ca.TestHubWinUI()
 	defer func() {
 		ui.SaveButton.SetText(saveButtonText)
 		ui.SaveButton.SetEnabled(true)
 	}()
 
 	if err != nil {
+		// do not set the error in the status bar by default
+		if err == cagent.ErrorTestWinUISettingsAreEmpty && testOnly {
+			return
+		}
+
 		ui.StatusBar.SetText("Status: failed to connect to the HUB ")
 		ui.StatusBar.SetIcon(ui.ErrorIcon)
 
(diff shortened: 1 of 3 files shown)


cmd/cagent/winui.go
```diff
@@ -143,7 +149,7 @@ func (ui *UI) TestSaveReload(testOnly bool) {
 
 	ui.SaveButton.SetText("Stopping the service...")
 	err = stopService(s)
-	if err != nil {
+	if err != nil && !strings.Contains(err.Error(), "has not been started") {
```
soupdiver: `has not been started` is correct?

cmd/cagent/winui.go
```diff
@@ -163,7 +169,7 @@ func (ui *UI) TestSaveReload(testOnly bool) {
 		}
 	}
 
-	RunDialog(ui.MainWindow, ui.SuccessIcon, "Success", message+"Service restarted and all changes applied!", callback)
+	RunDialog(ui.MainWindow, ui.SuccessIcon, "Success", message+"Services restarted and you are all set up!", callback)
```
soupdiver: Don't capitalise `Services`

handler.go
```diff
@@ -86,16 +88,58 @@ func (ca *Cagent) TestHub() error {
 	}
 
 	if resp.StatusCode == 401 && ca.Config.HubUser == "" {
-		return fmt.Errorf("unable to authorise without credentials. Please set hub_user & hub_password in the Config")
+		return fmt.Errorf("unable to authorise without credentials. Please set hub_user & hub_password in the Config according to your Cloudradar account")
 	} else if resp.StatusCode == 401 && ca.Config.HubUser != "" {
-		return fmt.Errorf("unable to authorise with the provided credentials. Please correct the hub_user & hub_password in the Config")
+		return fmt.Errorf("unable to authorise with provided credentials. Please correct the hub_user & hub_password in the Config according to your Cloudradar account")
```
soupdiver: Don't capitalise `Config`

handler.go
```diff
@@ -86,16 +88,58 @@ func (ca *Cagent) TestHub() error {
 	}
 
 	if resp.StatusCode == 401 && ca.Config.HubUser == "" {
-		return fmt.Errorf("unable to authorise without credentials. Please set hub_user & hub_password in the Config")
+		return fmt.Errorf("unable to authorise without credentials. Please set hub_user & hub_password in the Config according to your Cloudradar account")
```
soupdiver: Don't capitalise `Config`

cmd/cagent/winui.go
requilence: This starts from the new line like this:
image

cmd/cagent/winui.go
requilence: yes. This is error provided by windows service manager itself. That's why I'm not using the predefined error, just in case error it can slightly differ from system to system

PR #91, opened 2019-02-11
fix: hwinventory do not fail if DMI type not found

troian:
Some important DMI types like Base Board may be absent.
To fix continues inventory running even required fields are missing to gather as much information as possible

### DIFF
--- a/pkg/hwinfo/hwinfo_nonwindows.go
+++ b/pkg/hwinfo/hwinfo_nonwindows.go
@@ -44,7 +44,7 @@ func fetchInventory() (map[string]interface{}, error) {
 	res := make(map[string]interface{})
 
 	// all below requests are based on parsed data returned by dmidecode.Unmarshal
-	// refer to doc dmidecode.Get to get description of function behaviour
+	// refer to doc dmidecode.Get to get description of function behavior
 	var reqSys []dmidecode.ReqBaseBoard
 	if err = dmi.Get(&reqSys); err == nil {
 		res["baseboard.manufacturer"] = reqSys[0].Manufacturer
--- a/pkg/hwinfo/hwinfo_nonwindows.go
+++ b/pkg/hwinfo/hwinfo_nonwindows.go
@@ -9,6 +9,7 @@ import (
 	"os/exec"
 
 	"github.com/cloudradar-monitoring/dmidecode"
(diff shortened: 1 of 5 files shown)


pkg/hwinfo/hwinfo_nonwindows.go
```diff
@@ -41,46 +43,49 @@ func fetchInventory() (map[string]interface{}, error) {
 
 	res := make(map[string]interface{})
 
+	// all below requests are based on parsed data returned by dmidecode.Unmarshal
+	// refer to doc dmidecode.Get to get description of function behaviour
```
golangcibot: `behaviour` is a misspelling of `behavior` (from `misspell`)

pkg/hwinfo/hwinfo_nonwindows.go
```diff
@@ -9,13 +9,15 @@ import (
 	"os/exec"
 
 	"github.com/cloudradar-monitoring/dmidecode"
+	"github.com/sirupsen/logrus"
 )
 
 func isCommandAvailable(name string) bool {
-	cmd := exec.Command("/bin/sh", "-c", "command -v "+name)
+	cmd := exec.Command("/bin/sh", "-c", "command", "-v", name)
```
soupdiver: 👍

pkg/hwinfo/hwinfo_nonwindows.go
```diff
@@ -41,46 +43,49 @@ func fetchInventory() (map[string]interface{}, error) {
 
 	res := make(map[string]interface{})
 
+	// all below requests are based on parsed data returned by dmidecode.Unmarshal
+	// refer to doc dmidecode.Get to get description of function behavior
```
soupdiver: My intention was that you explain why you check against `err == nil` instead the "normal" `err != nil`
Explain what you explained to me in Slack, that `err == nil` means the value is present and `len() > 0`...

troian: `dmidecode.Get` has in description
```
// When function returns nil then dst contains slice with at least one valid entry
```
Do we need to copy?

soupdiver: yeah, make it as easy as possible to understand your code without the need to dig down the libraries

PR #92, opened 2019-02-12
Improve WMI timeouts handling

requilence:
- do not stop the CPU watcher if 1st query returns timeout
- change `cpuGetUtilisationTimeout` and `measureInterval` for 20 seconds
- return better human-readable error instead of `context deadline exceeded` for WMI queries in CPU and win processes list

### DIFF
--- a/cpu.go
+++ b/cpu.go
@@ -17,8 +17,8 @@ import (
 	log "github.com/sirupsen/logrus"
 )
 
-const measureInterval = time.Second * 20
-const cpuGetUtilisationTimeout = time.Second * 20
+const measureInterval = time.Second * 10
+const cpuGetUtilisationTimeout = time.Second * 10
 
 var errMetricsAreNotCollectedYet = errors.New("metrics are not collected yet")
 var utilisationMetricsByOS = map[string][]string{
@@ -273,6 +273,8 @@ func (ca *Cagent) CPUWatcher() *CPUWatcher {
 	// optimization to prevent CPU watcher to run in case CPU util metrics not are not needed
 	if len(ca.Config.CPUUtilTypes) > 0 && len(ca.Config.CPUUtilDataGather) > 0 || len(ca.Config.CPULoadDataGather) > 0 {
 		err := cw.Once()
+		// if err is nil or we got timeout error - we should run the CPU Watcher continuously
+		// in case we go some other kind of error we shouldn't start the CPU watcher because WMI appears disabled on the system
 		if _, timeoutError := err.(TimeoutError); err == nil || timeoutError {
 			go cw.Run()
 		}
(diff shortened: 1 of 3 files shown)


cpu.go
```diff
@@ -17,8 +17,8 @@ import (
 	log "github.com/sirupsen/logrus"
 )
 
-const measureInterval = time.Second * 10
-const cpuGetUtilisationTimeout = time.Second * 10
+const measureInterval = time.Second * 20
```
soupdiver: We can increase the timeouts but I don't think this will change anything with our current issue. I would guess it's something "deeper"

cpu.go
```diff
@@ -273,17 +273,23 @@ func (ca *Cagent) CPUWatcher() *CPUWatcher {
 	// optimization to prevent CPU watcher to run in case CPU util metrics not are not needed
 	if len(ca.Config.CPUUtilTypes) > 0 && len(ca.Config.CPUUtilDataGather) > 0 || len(ca.Config.CPULoadDataGather) > 0 {
 		err := cw.Once()
+		if _, timeoutError := err.(TimeoutError); err == nil || timeoutError {
```
soupdiver: I red this line 10 times and for me it's still not clear what the condition is here...
You check against `err == nil` and `err == timeoutError` in the same if branch. Please at least add a comment to explain what's going on here

cpu.go
```diff
@@ -293,6 +299,9 @@ func (cw *CPUWatcher) Once() error {
```
soupdiver: remove empty line

types.go
```diff
@@ -28,3 +33,12 @@ func floatToIntPercentRoundUP(f float64) int {
 func floatToIntRoundUP(f float64) int {
 	return int(f + 0.5)
 }
+
+type TimeoutError struct {
+	Origin  string
+	Timeout time.Duration
+}
+
+func (terr TimeoutError) Error() string {
```
soupdiver: Either use `timeoutError` or `err`

cpu.go
requilence: fixed

types.go
requilence: fixed

cpu.go
requilence: I split it and added the comment

soupdiver: ok, thanks for adding the comment.
Should we really run CPU Watcher continuously in case of timeout? After 3 or 5 timed out queries maybe we should stop and write a log statement?

cpu.go
requilence: done

cpu.go
requilence: first, let's try to fix this issue then we could try to optimise
I don't know, maybe there is a chance that WMI DLL can be locked for 1 minute so all 6 attempts with a 10-second timeout will fail

PR #93, opened 2019-02-12
fix instruction for build from sources

ochirovch:
after parameter -o must be filename
add $GOPATH in parameter git-dir for using relative path


soupdiver:
LGTM

PR #94, opened 2019-02-14
fix: check vmstat is available inside Acquire

troian:
Check hyper-v feature is available inside `Acquire`
Query WMI feature with timeout

### DIFF
--- a/handler.go
+++ b/handler.go
@@ -20,6 +20,7 @@ import (
 	"github.com/cloudradar-monitoring/cagent/pkg/monitoring/docker"
 	"github.com/cloudradar-monitoring/cagent/pkg/monitoring/services"
 	"github.com/cloudradar-monitoring/cagent/pkg/monitoring/vmstat"
+	vmstattypes "github.com/cloudradar-monitoring/cagent/pkg/monitoring/vmstat/types"
 )
 
 var ErrorTestWinUISettingsAreEmpty = errors.New("Please fill 'HUB URL', 'HUB USER' and 'HUB PASSWORD' from your Cloudradar account")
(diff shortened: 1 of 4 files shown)


handler.go
```diff
@@ -20,6 +20,7 @@ import (
 	"github.com/cloudradar-monitoring/cagent/pkg/monitoring/docker"
 	"github.com/cloudradar-monitoring/cagent/pkg/monitoring/services"
 	"github.com/cloudradar-monitoring/cagent/pkg/monitoring/vmstat"
+	"github.com/cloudradar-monitoring/cagent/pkg/monitoring/vmstat/types"
```

handler.go
```diff
@@ -387,12 +388,10 @@ func (ca *Cagent) getVMStatMeasurements(f func(string, MeasurementsMap, error))
 		for _, name := range ca.Config.VirtualMachinesStat {
 			vm, err := vmstat.Acquire(name)
 			if err != nil {
-				log.Warnf("acquire vm provider \"%s\": %s", name, err.Error())
-			} else {
-				if err = vm.IsAvailable(); err != nil {
-					log.Warnf("vm provider \"%s\" either not available or not enabled for host system %s: %s", name, runtime.GOOS, err.Error())
+				if err != vmstattypes.ErrNotAvailable {
+					log.Warnf("acquire vm provider \"%s\": %s", name, err.Error())
```
soupdiver: Change `acquire vm provide` into something more human-friendly?
`Error while acquiring vm provider or so`

Also, don't we usually have prefixes for the logs?

pkg/wmi/wmi.go
```diff
@@ -43,13 +43,13 @@ func (st FeatureInstallState) String() string {
 func CheckOptionalFeatureStatus(feature string) (FeatureInstallState, error) {
 	var dst []win32_optionalfeature
 	q := wmi.CreateQuery(&dst, "WHERE name = \""+feature+"\"")
-	err := wmi.Query(q, &dst)
+	err := QueryWithTimeout(10*time.Second, q, &dst)
```
soupdiver: make `10*time.Second` a `const`

handler.go

pkg/wmi/wmi.go
troian: done

handler.go

PR #99, opened 2019-02-21
feat: allow cagent running without output file and config

troian:
Cagent is not very friendly to automated installation because it requires a hub config for a clean installation which starts the agent as a service.
User can stuck if using script or Ansible. Config can be created prior to the installation of the package, but it requires all needed folders too. 

## Solution
If cagent is installed without a hub config injected via environment variables it should use a default config, which starts the daemon in local file mode.


requilence:
I don't fully get the reason. In my opinion, it's better to stop the installation than make it succeed and redirect to the `/dev/null`. Reasons:
- User see that service is successfully running, but in fact, it is in the failing state
- User needs to restart the service after he put the right config

PR #100, opened 2019-02-25
Improve CPU analysis and listing of 'top' consuming processes

nikita-vanyasin:
- gopsutil library updated to v2.19.01, which allowed simplifying some code in 'top' and 'cpu' modules
- WMI call eliminated when calculating CPU utilization (cpu.util.* fields in JSON). Now we have fewer timeouts and overall performance of cagent is improved.
- The listing of 'top' consuming process for Windows is changed. Now it uses Windows API call to ntdll.dll instead of Performance Counters. It allows getting the PID of processes. Also, in future, it will allow showing memory usage and other stats (they are already collected by this WinAPI)
- 'Load' field removed, as it seems useless to the user because it shows the last collected value (which can be very small, though the process may be the most consuming)
- For all operating systems: 'Load1', 'Load5', 'Load15' fields now has maximum value 100%, where 100% is full load of all cores.
- Removed unused libraries from 'vendor'
- Some refactorings


cpu_notwindows.go
```diff
@@ -0,0 +1,22 @@
+// +build !windows
+
+package cagent
+
+import (
+	"context"
+	"github.com/shirou/gopsutil/cpu"
```
golangcibot: File is not `goimports`-ed (from `goimports`)

pkg/winapi/winapi.go
```diff
@@ -0,0 +1,245 @@
...
+	maxBuffer := 2056
+	// buffer for results from the windows proc
+	resultBuffer := make([]SystemProcessorPerformanceInformation, maxBuffer)
+	// size of the buffer in memory
+	bufferSize := uintptr(systemProcessorPerformanceInfoSize) * uintptr(maxBuffer)
+	// size of the returned response
+	var retSize uint32
+
+	retCode, _, err := procNtQuerySystemInformation.Call(
+		systemProcessorPerformanceInformationClass, // System Information Class -> SystemProcessorPerformanceInformation
```
requilence: Windows docs suggest that you should no use `SystemProcessorPerformanceInformation` because it can differ from version to version and be discounted in the next version. They suggest using `GetSystemTimes` which is not per CPU :( 
https://docs.microsoft.com/en-us/windows/desktop/api/winternl/nf-winternl-ntquerysysteminformation#system_processor_performance_information

also: 
```
[NtQuerySystemInformation may be altered or unavailable in future versions of Windows. Applications should use the alternate functions listed in this topic.]
```

requilence: Is it ok to leave with it? Should we test on the each major version first?

nikita-vanyasin: > Is it ok to leave with it? Should we test on the each major version first?
Yeah, I've seen this warning regarding the usage of this winapi call.
In fact, any win api call can be altered in the future.

It's already tested on Win7,8,10 and WinServer 2012/2016/2019

Actually, the idea of using this call came from browsing signalfx/signalfx-agent implementation: https://github.com/signalfx/signalfx-agent/blob/e89f2c5fad499a2637010ac32a48b16a0d7b4173/internal/monitors/cpu/cpu_windows.go

handler.go
```diff
@@ -20,7 +20,7 @@ import (
 	"github.com/cloudradar-monitoring/cagent/pkg/monitoring/docker"
 	"github.com/cloudradar-monitoring/cagent/pkg/monitoring/services"
 	"github.com/cloudradar-monitoring/cagent/pkg/monitoring/vmstat"
-	vmstattypes "github.com/cloudradar-monitoring/cagent/pkg/monitoring/vmstat/types"
+	"github.com/cloudradar-monitoring/cagent/pkg/monitoring/vmstat/types"
```

nikita-vanyasin:
We decided to split these request onto three:
- vendor cleanup #102 
- cpu utilization #104
- hyperv types cleanup #103

PR #104, opened 2019-02-28
Improve cpu utilisation analysis

nikita-vanyasin:
* gopsutil library updated to v2.19.01, which allowed simplifying some code in 'top' and 'cpu' modules
* WMI call eliminated when calculating CPU utilization (cpu.util.* fields in JSON). Now we have fewer timeouts and overall performance of cagent is improved.
* The listing of 'top' consuming process for Windows is changed. Now it uses Windows API call to ntdll.dll instead of Performance Counters. It allows getting the PID of processes. Also, in future, it will allow showing memory usage and other stats (they are already collected by this WinAPI)
* 'Load' field removed, as it seems useless to the user because it shows the last collected value (which can be very small, though the process may be the most consuming)
* For all operating systems: 'Load1', 'Load5', 'Load15' fields now has maximum value 100%, where 100% is full load of all cores.
* some fields in JSON were renamed for consistency
* added information about CPU Utilisation Analysis configuration


handler.go
```diff
@@ -20,7 +20,7 @@ import (
 	"github.com/cloudradar-monitoring/cagent/pkg/monitoring/docker"
 	"github.com/cloudradar-monitoring/cagent/pkg/monitoring/services"
 	"github.com/cloudradar-monitoring/cagent/pkg/monitoring/vmstat"
-	vmstattypes "github.com/cloudradar-monitoring/cagent/pkg/monitoring/vmstat/types"
+	"github.com/cloudradar-monitoring/cagent/pkg/monitoring/vmstat/types"
```

requilence:
@troian feel free to metge if you are ok with it :)

PR #108, opened 2019-03-04
List open ports

nikita-vanyasin:
Cagent now will return a list of local open TCP ports similar to the output of "netstat -tulpen" on Linux.
An object  `"listeningport.list": []" is appended to the result object.

### DIFF
--- a/handler.go
+++ b/handler.go
@@ -232,13 +232,21 @@ func (ca *Cagent) GetAllMeasurements() (MeasurementsMap, error) {
 
 	measurements = measurements.AddWithPrefix("net.", netResults)
 
-	proc, err := ca.ProcessesResult()
+	proc, processList, err := ca.ProcessesResult()
 	if err != nil {
 		// no need to log because already done inside ProcessesResult()
 		errs = append(errs, err.Error())
 	}
 
 	measurements = measurements.AddWithPrefix("proc.", proc)
+
+	ports, err := ca.PortsResult(processList)
+	if err != nil {
+		// no need to log because already done inside PortsResult()
+		errs = append(errs, err.Error())
+	}
+
+	measurements = measurements.AddWithPrefix("listeningport.", ports)
 
 	mem, err := ca.MemResults()
 	if err != nil {
--- a/ports.go
+++ b/ports.go
@@ -1,0 +1,78 @@
+package cagent
+
+import (
+	"fmt"
+	"github.com/shirou/gopsutil/net"
+	log "github.com/sirupsen/logrus"
+	"syscall"
+)
+
+type PortStat struct {
+	Protocol     string `json:"proto"`
+	LocalAddress string `json:"addr"`
+	State        string `json:"state,omitempty"`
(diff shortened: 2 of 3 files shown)


ports.go
```diff
@@ -0,0 +1,73 @@
+package cagent
+
+import (
+	"fmt"
```
golangcibot: File is not `goimports`-ed (from `goimports`)

handler.go
```diff
@@ -232,14 +232,22 @@ func (ca *Cagent) GetAllMeasurements() (MeasurementsMap, error) {
...
 	if err != nil {
 		// no need to log because already done inside ProcessesResult()
 		errs = append(errs, err.Error())
 	}
 
 	measurements = measurements.AddWithPrefix("proc.", proc)
 
+	ports, err := ca.PortsResult(processList)
+	if err != nil {
+		// no need to log because already done inside PortsResult()
```
soupdiver: Do the error logging here, see comment inside PortsResult for explanation

ports.go
```diff
@@ -0,0 +1,74 @@
...
+	Protocol     string `json:"proto"`
+	LocalAddress string `json:"addr"`
+	PID          int32  `json:"pid,omitempty"`
+	ProgramName  string `json:"program,omitempty"`
+}
+
+func (ca *Cagent) PortsResult(processList []ProcStat) (m MeasurementsMap, err error) {
+	connections, err := net.Connections("inet")
+	if err != nil {
+		log.Error("[PORTS] error: ", err.Error())
```
soupdiver: Logging an error is basically handling it. Return the error without logging here and let the caller of this function log the error

ports.go
```diff
@@ -0,0 +1,74 @@
...
+		return nil, err
+	}
+
+	var ports []PortStat
+	for _, conn := range connections {
+		state := conn.Status
+		if state != "LISTEN" && state != "NONE" && state != "" {
```
soupdiver: Please add a quick comment _why_ we continue in these cases

ports.go
```diff
@@ -0,0 +1,74 @@
...
+
+		ports = append(ports, PortStat{
+			Protocol:     formatConnectionProtocol(conn.Family, conn.Type),
+			LocalAddress: formatNetAddr(&conn.Laddr),
+			PID:          conn.Pid,
+			ProgramName:  programName,
+		})
+	}
+
+	log.Info("[PORTS] results: ", len(ports))
```
soupdiver: Is this log.Info or log.Debug?

ports.go
nikita-vanyasin: caller is handler.GetAllMeasurements() 
This function does not do any logging at all. It's just checking if there were any errors.

I agree that the caller should do the logging, not the child, but I think it's better to implement it as separate PR for all calls from within GetAllMeasurements() function.

handler.go
soupdiver: Will be handled in separate PR

PR #111, opened 2019-03-12
Eliminate wmi query for process list

nikita-vanyasin:
WMI query is slow and actually does not work when the system is under high load.

1) This PR replaces WMI query with winapi calls, which are:
- list all the processes in system using one winapi call
- for each process tries to get its command line string. (3-4 winapi calls). 
2) Also, some refactorings made to simplify winapi module.

***
Note that in some cases we are unable to get the command line string at all:
- for some system processes, which memory is inaccessible in user mode
- when running 32-bit cagent under WOW64, only 32-bit processes memory is accessible for cagent.
***
Tested on Win10, Windows Server (2012, 2016, 2019)

### DIFF
--- a/pkg/winapi/winapi.go
+++ b/pkg/winapi/winapi.go
@@ -183,12 +183,11 @@ func GetProcessBasicInformation(processHandle windows.Handle) (*processBasicInfo
 }
 
 func ReadProcessMemory(processHandle windows.Handle, srcAddr uintptr, dstAddr uintptr, size uintptr) (int, error) {
-	var nBytesRead int
-
 	if err := checkKernel32ProceduresAvailable(); err != nil {
-		return nBytesRead, err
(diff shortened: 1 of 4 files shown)


pkg/winapi/winapi.go
```diff
@@ -243,3 +158,49 @@ func GetSystemProcessInformation() (map[uint32]*SystemProcessInformation, error)
...
+	}
+
+	return &pbi, nil
+}
+
+func ReadProcessMemory(processHandle windows.Handle, srcAddr uintptr, dstAddr uintptr, size uintptr) (int, error) {
+	var nBytesRead int
+
+	if err := checkKernel32ProceduresAvailable(); err != nil {
+		return nBytesRead, err
```
soupdiver: return explicit 0 in error case

PR #115, opened 2019-03-13
Eliminate WMI call for CPU name retrieval on Windows

nikita-vanyasin:
Now it will use value from registry.

### DIFF
--- a/system_windows.go
+++ b/system_windows.go
@@ -8,18 +8,18 @@ import (
 )
 
 func getProcessorModelName() (string, error) {
-	k, err := registry.OpenKey(registry.LOCAL_MACHINE, `HARDWARE\DESCRIPTION\System\CentralProcessor\0`, registry.QUERY_VALUE)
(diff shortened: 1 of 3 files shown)


system_windows.go
```diff
@@ -0,0 +1,28 @@
...
+
+package cagent
+
+import (
+	log "github.com/sirupsen/logrus"
+	"golang.org/x/sys/windows/registry"
+)
+
+func getProcessorModelName() (string, error) {
+	k, err := registry.OpenKey(registry.LOCAL_MACHINE, `HARDWARE\DESCRIPTION\System\CentralProcessor\0`, registry.QUERY_VALUE)
```
soupdiver: Please use a more descriptive name than `k`

nikita-vanyasin: Although Golang coding guidelines recommend using short variables names when the scope is small, I love descriptive variable names too.   Fixed :+1:

soupdiver: Short is good but not _too_ short :)

PR #116, opened 2019-03-20
Resolve "error LGHT0204 : ICE64"

xlab:
Also add build script for MinGW64 environments.

### DIFF
--- a/pkg-scripts/msi-templates/product.wxs
+++ b/pkg-scripts/msi-templates/product.wxs
@@ -70,6 +70,7 @@
                 <Directory Id="INSTALLDIR" Name="{{.Product}}">
                     {{if gt (.Files.Items | len) 0}}
                     <Component Id="ApplicationFiles" Guid="{{.Files.GUID}}">
+                        <RemoveFolder Id="CleanApplicationFolders" Directory="ApplicationProgramsFolder" On="uninstall"/>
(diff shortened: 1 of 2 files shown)


pkg-scripts/msi-templates/product.wxs
```diff
@@ -70,6 +70,7 @@
                 <Directory Id="INSTALLDIR" Name="{{.Product}}">
                     {{if gt (.Files.Items | len) 0}}
                     <Component Id="ApplicationFiles" Guid="{{.Files.GUID}}">
+                        <RemoveFolder Id="CleanApplicationFolders" Directory="ApplicationProgramsFolder" On="uninstall"/>
```
nikita-vanyasin: Not sure if this statement will remove the old config and log files that are stored in this folder by default? I think they must be saved after uninstall.

xlab: Old config and log files are living inside `INSTALLDIR` inside Program Files. So it's another flow.

This action marks the `Cagent` directory from Start Menu for removal, it is defined above:

```xml
<Directory Id="ProgramMenuFolder">
       <Directory Id="ApplicationProgramsFolder" Name="Cagent"/>
</Directory>
```

I tested MSI from both branches and their behaviour is the same. This change only dismisses "error LGHT0204 : ICE64" warning.

PR #117, opened 2019-03-21
Extend hardware discovery with PCI, USB and Display

nikita-vanyasin:
see DEV-739.

Main changes:
- pkg/hwinfo package slightly refactored to support extension
- Implemented listing of PCI, USB, and monitors for linux-based, Windows and OS/X
- Some libraries were added to support the implementation
- pkg/common package was added with only class ErrorCollector which allows accumulating errors. This class simplifies the case when we don't need to return from a function in the case of error.


pkg/hwinfo/hwinfo_nonwindows.go
```diff
@@ -8,8 +8,8 @@ import (
 	"fmt"
 	"os/exec"
 
+	"github.com/cloudradar-monitoring/cagent/pkg/common"
```
golangcibot: File is not `goimports`-ed (from `goimports`)

pkg/hwinfo/hwinfo_other.go
```diff
@@ -0,0 +1,197 @@
...
+	"fmt"
+	"io"
+	"io/ioutil"
+	"os"
+	"os/exec"
+	"regexp"
+	"strconv"
+	"strings"
+
+	"github.com/cloudradar-monitoring/cagent/pkg/common"
```
golangcibot: File is not `goimports`-ed (from `goimports`)

pkg/hwinfo/system_profiler.go
```diff
@@ -0,0 +1,166 @@
+// Package for parsing the XML output of system_profiler command output (available on OS X)
+
+package hwinfo
+
+import (
+	"fmt"
+	"io"
+	"strings"
+
+	"github.com/cloudradar-monitoring/cagent/pkg/common"
```
golangcibot: File is not `goimports`-ed (from `goimports`)

Gopkg.toml
```diff
@@ -68,3 +68,19 @@
 [[constraint]]
   branch = "master"
   name = "github.com/lxn/walk"
+
+[[constraint]]
+  name = "github.com/jaypipes/ghw"
+  version = "0.3.0"
+
+[[constraint]]
+  branch = "master"
```
soupdiver: Shouldn't we ping to a version or commit instead simply `master`?

Gopkg.toml
```diff
@@ -68,3 +68,19 @@
...
+  name = "github.com/vcraescu/go-xrandr"
+
+[[constraint]]
+  name = "github.com/gentlemanautomaton/windevice"
+  revision = "de21ffdab1a3b60205fe0782a9df61feb34367f1"
+
+[[constraint]]
+  branch = "master"
```
soupdiver: Shouldn't we ping to a version or commit instead simply `master`?

pkg/hwinfo/hwinfo.go
```diff
@@ -1,16 +1,48 @@
...
-		return nil, err
+		err = errors.Wrap(err, "hwinfo")
+		log.Error(err)
+		return hw, err
 	}
 
 	return hw, nil
 }
+
+func mergeStringMaps(mapA, mapB *map[string]interface{}) map[string]interface{} {
```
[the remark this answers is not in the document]
xlab: In Go it's common to have simple private helpers, this thing contains a simple `for {}` loop, why move it to another package, also its effect can be seen here — what exactly it does. Otherwise it would require a package lookup.
See [Proverb No.8](https://go-proverbs.github.io)

pkg/hwinfo/hwinfo.go
```diff
@@ -1,16 +1,48 @@
...
+	result := *mapA
```
xlab: Not sure why this is here instead just map-by-value. It wouldn't create a copy anyway.
https://play.golang.org/p/9bFI7agWcJ3

pkg/hwinfo/hwinfo_nonwindows.go
```diff
@@ -22,9 +22,36 @@ func isCommandAvailable(name string) bool {
...
+
+	dmiDecodeResults := retrieveInfoUsingDmiDecode(&errorCollector)
+	if len(dmiDecodeResults) > 0 {
+		res = mergeStringMaps(&res, &dmiDecodeResults)
+	}
+
+	return res, errorCollector.Combine()
+}
+
+func retrieveInfoUsingDmiDecode(errs *common.ErrorCollector) map[string]interface{} {
```
xlab: Can we stick to idiomatic Go function patterns like `func foo() error`. I don't like that a function related to business logic of the app has so much noise in its Inputs and Outputs. Both `*common.ErrorCollector `and `map[string]interface{}` has nothing to do with function logic, they are orchestration-scope helpers.

You can create an idiomatic function like `func foo() error` and then collect the errors in the orchestrator itself, e.g.
```go
var err error
var errs *common.ErrorCollector

err = foo()
errs.Add(err)
err = bar()
errs.Add(err)

errorCollector.Combine()
```

Also, there is a package called https://github.com/pkg/errors that is designed to provide multi-error capabilities without breaking the flow.

pkg/hwinfo/hwinfo_nonwindows.go
```diff
@@ -51,14 +80,14 @@ func fetchInventory() (map[string]interface{}, error) {
 		res["baseboard.model"] = reqSys[0].Version
 		res["baseboard.serial_number"] = reqSys[0].SerialNumber
 	} else if err != dmidecode.ErrNotFound {
-		logrus.Errorf("hwinfo: failed fetching baseboard info, %s", err.Error())
+		errs.Addf("failed fetching baseboard info, %s", err.Error())
```
xlab: Error context gets missing here (e.g. error origin - `hwinfo:`). Maybe leverage https://github.com/pkg/errors to wrap errors with context and return them all in stack?

pkg/hwinfo/hwinfo_other.go
```diff
@@ -0,0 +1,197 @@
...
+
+	return buf.String(), nil
+}
+
+func listPCIDevices(errs *common.ErrorCollector) []*pciDeviceInfo {
+	var ghwErr error
+	var devices []*ghw.PCIDevice
+
+	// unfortunately, jaypipes/ghw library sometime writes error message directly to os.Stderr instead of returning from function call.
+	// that's why we will try to capture stderr output and handle it:
```
xlab: Maybe vendor it with some fix inside? The stderr capturing looks too hacky.

pkg/hwinfo/hwinfo_other.go
```diff
@@ -0,0 +1,197 @@
...
+		}
+		devNum, err := strconv.Atoi(reg.FindString(sanitizedTokens[3]))
+		if err != nil {
+			errs.Addf("error while parsing device number: %s. line: %s", err.Error(), lines[i])
+			continue
+		}
+		address := fmt.Sprintf("bus %d device %d", busNum, devNum)
+		devID := lsusbLineRegexp.FindString(sanitizedTokens[5])
+		results = append(results, &usbDeviceInfo{
+			address,
```
xlab: Please always used named struct literals, e.g. with field names. It's extremely hard to understand, expand and maintain such declarations, without field names.

pkg/hwinfo/hwinfo_windows.go
```diff
@@ -7,127 +7,226 @@ import (
...
+			errs.Addf("could not get USB device class %s", err.Error())
+		}
+		if class == "USB" {
+			class = ""
+		}
+
+		descriptionParts := []string{fullDisplayingName, friendlyName, class}
+		description := ""
+		for _, part := range descriptionParts {
+			if part != "" {
```
xlab: Please use `len(x) == 0` and `len(x) > 0` instead of `x == ""` and `x != ""`.

pkg/hwinfo/hwinfo.go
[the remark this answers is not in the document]
soupdiver: I just expressed my opinion. I don't disagree to have simple private helpers. I _personally_ think that this helper is so generic, that there is no need to have it as a _private_ function inside this package.
I agree that some copying is often better than adding an _external_ dependency but this lies entirely within our codebase.
Different opinions welcome :)

pkg/hwinfo/hwinfo_nonwindows.go
soupdiver: > Error context gets missing here
yip, this is a problem over the whole codebase.
I have this in my mind for some time. Maybe this can be addressed with some focused refactor effort.

pkg/hwinfo/hwinfo_other.go
soupdiver: > The stderr capturing looks too hacky

I agree but vendoring it with a fix, forces us to "maintain" that fork, right?

pkg/hwinfo/hwinfo.go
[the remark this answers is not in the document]
xlab: @soupdiver sorry if my comment seemed harsh, from my history of legacy codebases, moving primitive functions into packages like "common" was never a good decision - it still get used only once, and there can be multiple packages called "util" and "common" in the end, with functions like left-pad or isArray (node.js pun intended). So we've got different views on this, haha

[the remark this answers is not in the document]
soupdiver: Fair reasoning
I can live with that argumentation :)

> sorry if my comment seemed harsh

all good 😁

pkg/hwinfo/hwinfo_other.go
xlab: It depends really, I don't know how frequently this particular package is updated. And we'll be vendoring it anyway. Usually task can be simplified by creating a patch, so it could be applied when you bump vendored version of the target package. I still believe it's more worthy than hijacking Stderr.

soupdiver: > I don't know how frequently this particular package is being updated. 

Have a look at the recent commits, there is some recent activity.
Having a patch and apply it over and over again introduces a new workflow which we don't have at the moment. So I'm not a huge fan of this.
Having this hack, at least keep it entirely in our codebase and we still get updates from upstream the normal way.

pkg/hwinfo/hwinfo_nonwindows.go
nikita-vanyasin: Wrap(err, "hwinfo") was moved one abstract level higher to hwinfo.go. This is why "hwinfo:" was removed from this place.

pkg/hwinfo/hwinfo_windows.go
nikita-vanyasin: Not sure why? Both ways are idiomatic and used in the standard libraries. In both cases, the compiler output should be the same.

pkg/common/errors.go
```diff
@@ -0,0 +1,45 @@
...
+import (
+	"errors"
+	"fmt"
+)
+
+type ErrorCollector struct {
+	errs []error
+}
+
+func (c *ErrorCollector) New(err error) {
```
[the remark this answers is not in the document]
nikita-vanyasin: I believe that code within function declaration should be self-documenting. Please specify which loop logins is unclear.

pkg/hwinfo/hwinfo_nonwindows.go
nikita-vanyasin: I agree, that ErrorCollector looks like too much for most of the simple functions, but in complex functions like `hwinfo_other.go: listUSBDevices`, I could not find a better way to collect errors from inside of the loop. In other places, ErrorCollector is used only for consistency.

map[string]interface{} is a piece of old hwinfo implementation that I decided to leave intact in this PR, as it's not directly related to the task. In the future, we can rework it in a separate PR (for example returning meaningful types instead of string maps).

Not sure what do you mean by 'orchestration' in this context, though.
Also, not sure how `github.com/pkg/errors` can help when working with collecting errors.

pkg/common/errors.go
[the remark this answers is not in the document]
nikita-vanyasin: Regarding the ErrorCollector: renamed to Add, AddNew and AddNewf

pkg/common/errors.go
```diff
@@ -0,0 +1,45 @@
+package common
+
+import (
+	"errors"
+	"fmt"
+)
+
+type ErrorCollector struct {
```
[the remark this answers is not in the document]
nikita-vanyasin: done

pkg/hwinfo/hwinfo_other.go
```diff
@@ -0,0 +1,197 @@
...
+
+func captureStderr(funcToExecute func()) (string, error) {
+	r, w, err := os.Pipe()
+	if err != nil {
+		return "", err
+	}
+
+	stderr := os.Stderr
+	os.Stderr = w
+	defer func() {
```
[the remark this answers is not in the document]
nikita-vanyasin: done

Gopkg.toml
nikita-vanyasin: done

Gopkg.toml
nikita-vanyasin: done

pkg/hwinfo/hwinfo.go
nikita-vanyasin: fixed

pkg/hwinfo/hwinfo_other.go
nikita-vanyasin: Actually, I used to the fact that my IDE show it automatically:
image

Fixed.

pkg/hwinfo/hwinfo_nonwindows.go
xlab: I agree, let't have this as a different task, it definitely requires a single big refactor targeting error collection & error contexts. As for 'orchestration', it named the logic that will call these functions and collect their results.

pkg/hwinfo/hwinfo_windows.go
xlab: Yes, but the first one is agnostic to the type. If the value going to change from `string` to `[]byte` the code will stay the same. I know that in this particular case it probably will never change, but anyways. These types are interchangeable, according to Go spec they mean the same:
```
A slice is a descriptor for a contiguous segment of an underlying array and provides access to a numbered sequence of elements from that array.

A string value is a (possibly empty) sequence of bytes. 
```

Check like `len(x) > 0 ` is easy to read, because it literally reads "length more than zero", the comparison `x != ""` looks like you are trying to find an exact value (of exact type). Moreover, sometimes the check needs to be adjusted, for example, skip not only empty strings, but short ones too. If you have `len(x) > 0` this is one digit to change: `len(x) > 2`, but in case of comparison `x != ""` you'll have to rewrite the whole line. Switching back will be painful too.

I'm okay if this change will be unaddressed in this PR, to avoid extra work, but is is something to consider for the future code and refactoring. ;)

pkg/common/errors.go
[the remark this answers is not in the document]
soupdiver: I don't think `AddNew` is much clearer than `Add`. It's the same. _What_ is added?
So would prefer something like `AddErrorFromString`.
@xlab you have a take on that one?

Maybe it's just me 😅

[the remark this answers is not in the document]
soupdiver: See https://github.com/cloudradar-monitoring/cagent/pull/117#pullrequestreview-217798670

pkg/common/errors.go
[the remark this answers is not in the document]
soupdiver: and for `ErrorCollector` as well. Forgot to mention this explicitly

nikita-vanyasin:
> For example: https://github.com/cloudradar-monitoring/cagent/pull/117/files#diff-b76cb11727fb6ed8fdea4a8d240fbc37R133

I've simplified the code and added some comment. Are there any other loops that you want to be commented?

handler.go
```diff
@@ -274,11 +274,11 @@ func (ca *Cagent) GetAllMeasurements() (MeasurementsMap, error) {
...
 
 	ca.hwInventory.Do(func() {
-		if hwInfo, err := hwinfo.Inventory(); err != nil && err != hwinfo.ErrNotPresent {
+		hwInfo, err := hwinfo.Inventory()
+		if err != nil {
 			errs = append(errs, err.Error())
-		} else if err == nil {
-			measurements = measurements.AddInnerWithPrefix("hw.inventory", hwInfo)
 		}
+		measurements = measurements.AddInnerWithPrefix("hw.inventory", hwInfo)
```
soupdiver: We also add `hwInfo` in error case?

pkg/hwinfo/hwinfo.go
```diff
@@ -1,15 +1,39 @@
...
+	Size        string `json:"size,omitempty"`
+	Resolution  string `json:"resolution,omitempty"`
+}
 
 func Inventory() (map[string]interface{}, error) {
 	hw, err := fetchInventory()
 	if err != nil {
-		return nil, err
+		err = errors.Wrap(err, "hwinfo")
+		log.Error(err)
```
soupdiver: Is it correct that we return _and_ log the error? Shouldn't the logging be done when the error is handled?

pkg/hwinfo/hwinfo_nonwindows.go
xlab: Sorry I mistook the `github.com/pkg/errors` package with https://github.com/hashicorp/go-multierror however `pkg/errors` is cool too. They can be combined for best results.

soupdiver: I agree that `pkg/errors` is neat. I use it in every of my side-projects. But I would favour an approach where we update the whole codebase at once using that package.
Should be a lot of changes but since the logic doesn't change it should be doable in a (very) short amount of time

handler.go
nikita-vanyasin: Yes. This function can return partial result (as many others that are called in handler.GetMeasures()). For example, it can retrieve the list of USB devices but not the information about CPU (because the dmidecode is not installed). We should render as much info as possible in the result JSON, so we try. It's actually old code slightly refactored. I just made it more explicit.

pkg/hwinfo/hwinfo.go
nikita-vanyasin: Yes. We have already discussed this in some other PR before. The reason is the caller of this function (`handler.GetAllMeasurements()`) does not have any error logging code. This function now is similar to other functions called from GetAllMeasurements().

In this PR I just moved ALL logging from low-level implementation here, making it's more explicit and eliminating the code duplication.

soupdiver:
Can this be merged or is anything left from your side @xlab ?

PR #119, opened 2019-03-25
Refactor error representations, plus winui hub config refactorings.

xlab:
* Refactor error representations (added pkg/wrap, proper logrus fields);
* Refactor winui hub config interaction.

This PR targets better error handling in cagent, also adds more control for timeouts and error cascading.

### DIFF
--- a/cmd/cagent/main.go
+++ b/cmd/cagent/main.go
@@ -478,7 +478,7 @@ func runUnderOsServiceManager(ca *cagent.Cagent) {
 }
 
 func writePidFileIfNeeded(ca *cagent.Cagent, oneRunOnlyModePtr *bool) {
-	if ca.Config.PidFile != "" && !*oneRunOnlyModePtr && runtime.GOOS != "windows" {
+	if len(ca.Config.PidFile) > 0 && !*oneRunOnlyModePtr && runtime.GOOS != "windows" {
 		err := ioutil.WriteFile(ca.Config.PidFile, []byte(strconv.Itoa(os.Getpid())), 0664)
 		if err != nil {
 			log.WithFields(log.Fields{
@@ -489,11 +489,12 @@ func writePidFileIfNeeded(ca *cagent.Cagent, oneRunOnlyModePtr *bool) {
 }
 
 func removePidFileIfNeeded(ca *cagent.Cagent, oneRunOnlyModePtr *bool) {
-	if ca.Config.PidFile != "" && !*oneRunOnlyModePtr && runtime.GOOS != "windows" {
-		err := os.Remove(ca.Config.PidFile)
-		log.WithFields(log.Fields{
-			"pidFile": ca.Config.PidFile,
-		}).WithError(err).Errorf("Failed to remove pid file")
+	if len(ca.Config.PidFile) > 0 && !*oneRunOnlyModePtr && runtime.GOOS != "windows" {
+		if err := os.Remove(ca.Config.PidFile); err != nil {
+			log.WithFields(log.Fields{
+				"pidFile": ca.Config.PidFile,
+			}).WithError(err).Errorf("Failed to remove pid file")
+		}
 	}
 }
 
--- a/cmd/cagent/main.go
+++ b/cmd/cagent/main.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"bufio"
+	"context"
 	"flag"
 	"fmt"
 	"io/ioutil"
@@ -16,6 +17,7 @@ import (
 	"syscall"
 
 	"github.com/kardianos/service"
+	"github.com/pkg/errors"
 	log "github.com/sirupsen/logrus"
 
 	"github.com/cloudradar-monitoring/cagent"
@@ -41,7 +43,7 @@ func askForConfirmation(s string) bool {
 
 		response, err := reader.ReadString('\n')
 		if err != nil {
-			log.Fatalf("Failed to read confirmation: %s", err.Error())
+			log.WithError(err).Fatalln("Failed to read confirmation")
 		}
 
 		response = strings.ToLower(strings.TrimSpace(response))
@@ -97,19 +99,17 @@ func main() {
 	if serviceInstallUserPtr != nil && *serviceInstallUserPtr != "" ||
 		serviceInstallPtr != nil && *serviceInstallPtr {
 		if *outputFilePtr != "" {
-			fmt.Println("Output file(-o) flag can't be used together with service install(-s) flag")
-			os.Exit(1)
+			log.Fatalln("Output file(-o) flag can't be used together with service install(-s) flag")
 		}
 
 		if *serviceUninstallPtr {
-			fmt.Println("Service uninstall(-u) flag can't be used together with service install(-s) flag")
-			os.Exit(1)
+			log.Fatalln("Service uninstall(-u) flag can't be used together with service install(-s) flag")
 		}
 	}
 
 	cfg, err := cagent.HandleAllConfigSetup(*cfgPathPtr)
 	if err != nil {
-		log.Fatalf("Failed to handle cagent configuration: %s", err.Error())
+		log.WithError(err).Fatalln("Failed to handle Cagent configuration")
 	}
 
 	ca := cagent.New(cfg, *cfgPathPtr, version)
@@ -117,7 +117,7 @@ func main() {
 	handleFlagPrintConfig(*printConfigPtr, cfg)
 
 	if ((serviceInstallPtr == nil) || ((serviceInstallPtr != nil) && (!*serviceInstallPtr))) &&
-		((serviceInstallUserPtr == nil) || ((serviceInstallUserPtr != nil) && (*serviceInstallUserPtr == ""))) &&
+		((serviceInstallUserPtr == nil) || ((serviceInstallUserPtr != nil) && len(*serviceInstallUserPtr) == 0)) &&
 		!*serviceUninstallPtr {
 		handleServiceCommand(ca, *flagServiceStatusPtr, *flagServiceStartPtr, *flagServiceStopPtr, *flagServiceRestartPtr)
 	}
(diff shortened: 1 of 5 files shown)


handler.go
```diff
@@ -23,170 +26,160 @@ import (
...
-
-	req.Header.Add("User-Agent", ca.userAgent())
-	if ca.Config.HubUser != "" {
-		req.SetBasicAuth(ca.Config.HubUser, ca.Config.HubPassword)
-	}
-
-	resp, err := ca.hubHTTPClient.Do(req)
-	if err != nil {
-		return fmt.Errorf("Unable to connect. %s. If you have a proxy or firewall, it may be blocking the connection", err.Error())
+	io.Copy(ioutil.Discard, resp.Body)
```
golangcibot: Error return value of `io.Copy` is not checked (from `errcheck`)

xlab:
Files changed:

* [cagent.go](https://github.com/cloudradar-monitoring/cagent/blob/cf7e36c31bbf6012d8f912ad2c282f3e4b8492e0/cagent.go)
* [cmd/cagent/main.go](https://github.com/cloudradar-monitoring/cagent/blob/cf7e36c31bbf6012d8f912ad2c282f3e4b8492e0/cmd/cagent/main.go)
* [cmd/cagent/winui.go](https://github.com/cloudradar-monitoring/cagent/blob/cf7e36c31bbf6012d8f912ad2c282f3e4b8492e0/cmd/cagent/winui.go)
* [handler.go](https://github.com/cloudradar-monitoring/cagent/blob/cf7e36c31bbf6012d8f912ad2c282f3e4b8492e0/handler.go)

Pro tip: it's better to review this PR using editor and read each file as whole, comparing them side by side. It's hard to get the concept studying lines in github diff.

cmd/cagent/main.go
```diff
@@ -481,17 +481,19 @@ func writePidFileIfNeeded(ca *cagent.Cagent, oneRunOnlyModePtr *bool) {
...
 	}
 }
 
 func removePidFileIfNeeded(ca *cagent.Cagent, oneRunOnlyModePtr *bool) {
 	if ca.Config.PidFile != "" && !*oneRunOnlyModePtr && runtime.GOOS != "windows" {
 		err := os.Remove(ca.Config.PidFile)
-		if err != nil {
-			log.Errorf("Failed to remove pid file at: %s", ca.Config.PidFile)
-		}
+		log.WithFields(log.Fields{
```
nikita-vanyasin: Not sure why the condition was removed:
`if err != nil {`

handler.go
```diff
@@ -23,170 +26,159 @@ import (
...
+		}
 		if ca.rootCAs != nil {
-			tr.TLSClientConfig = &tls.Config{RootCAs: ca.rootCAs}
+			transport.TLSClientConfig = &tls.Config{
+				RootCAs: ca.rootCAs,
+			}
 		}
-		if ca.Config.HubProxy != "" {
+		if len(ca.Config.HubProxy) > 0 {
+			// TODO(max): this should be moved to config sanitizer/validator layer.
```
nikita-vanyasin: Do we have issue added to the tracking system for this?
I think leaving `TODO:` comments without clear plans when it will be fixed is not ok.

soupdiver: > Do we have issue added to the tracking system for this?

I don't think so

handler.go
```diff
@@ -23,170 +26,159 @@ import (
...
+		}
+		if err := gzipped.Close(); err != nil {
+			err = errors.Wrap(err, "failed to finalize gzipped buffer")
+			return err
+		}
+		req, _ = http.NewRequest("POST", ca.Config.HubURL, buf)
 		req.Header.Set("Content-Encoding", "gzip")
 	} else {
-		req, err = http.NewRequest("POST", ca.Config.HubURL, bytes.NewBuffer(b))
+		req, _ = http.NewRequest("POST", ca.Config.HubURL, bytes.NewBuffer(b))
```
soupdiver: Can we ignore this `err`?

handler.go
```diff
@@ -23,170 +26,159 @@ import (
...
 		req.SetBasicAuth(ca.Config.HubUser, ca.Config.HubPassword)
 	}
-
-	resp, err := ca.hubHTTPClient.Do(req)
-
-	if err != nil {
+	req = req.WithContext(ctx)
+	resp, err := ca.hubClient.Do(req)
+	if err = ca.checkClientError(resp, err, "hub_user", "hub_password"); err != nil {
 		return err
```
soupdiver: Shouldn't we wrap the error?

handler.go
```diff
@@ -342,25 +334,26 @@ func (ca *Cagent) GetAllMeasurements() (MeasurementsMap, error) {
...
-			return fmt.Errorf("Results json encode error: %s", err.Error())
+			err = errors.Wrap(err, "failed to JSON encode measurement result")
+			return err
 		}
-
 		return nil
 	}
-
-	err := ca.PostResultsToHub(result)
+	ctx, cancelFn := context.WithTimeout(context.Background(), 5*time.Minute)
```
soupdiver: Do `defer cancelFn()` instead calling it in every path?

cmd/cagent/main.go
xlab: That was typo, thank you for vigilance.

cmd/cagent/main.go
```diff
@@ -97,27 +99,25 @@ func main() {
...
+		log.WithError(err).Fatalln("Failed to handle Cagent configuration")
 	}
 
 	ca := cagent.New(cfg, *cfgPathPtr, version)
 
 	handleFlagPrintConfig(*printConfigPtr, cfg)
 
 	if ((serviceInstallPtr == nil) || ((serviceInstallPtr != nil) && (!*serviceInstallPtr))) &&
-		((serviceInstallUserPtr == nil) || ((serviceInstallUserPtr != nil) && (*serviceInstallUserPtr == ""))) &&
+		((serviceInstallUserPtr == nil) || ((serviceInstallUserPtr != nil) && len(*serviceInstallUserPtr) == 0)) &&
```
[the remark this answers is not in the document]
xlab: https://github.com/cloudradar-monitoring/cagent/pull/117#discussion_r267868494

I think there is an advantage. I prefer to use `len()` everywhere. The problem, in short, that Go is statically typed, so once you change the type, you'll need to rewrite a lot of code, that could stay the same if written using `len()` initially..

handler.go
xlab: `NewRequest` may return error in these cases:
* Invalid method (it's hardcoded here)
* Invalid URL (but it's checked in the code above)

So given that both parameters are valid, there will be no expected error.

handler.go
xlab: `checkClientError` is the wrapper for error from `hubClient.Do(req)`. It will parse it and response and format in a proper way. You suggest that a wrapper here could capture the stack and that could be useful later?

cmd/cagent/main.go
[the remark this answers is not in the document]
nikita-vanyasin: Well, I think this is okay for new code, but I don't see any point changing the already written checks. It's not such a critical problem :) 
I think we can leave it as it is in current PR, but in future avoid modifying checks just for the sake of modifying.

handler.go
soupdiver: > So given that both parameters are valid, there will be no expected error.

Well, I would still prefer an explicit error checking... who knows how the code might be changed in the future.

cmd/cagent/main.go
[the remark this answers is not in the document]
soupdiver: Please let the person who opened a question resolve it.
Otherwise it's easily missed what's the explanation.

> but in future avoid modifying checks just for the sake of modifying.

I agree

>  The problem, in short, is that Go is statically typed, so once you change the type, you'll need to rewrite a lot of code

I don't think this is a problem, even if it can be a bit annoying. If the type changes, I think it's good to be forced to have a look at the implications and not mask this behind the use of `len()`

handler.go
soupdiver: @xlab Please don't simply resolve the conversation, especially when you ask another question in the.

> You suggest that a wrapper here could capture the stack and that could be useful later?

From my personal experience, it can be very helpful to _always_ wrap an error. In case of debugging, it can safe a lot of time to immediately have the full stack. Also it's an easy pattern to follow. If you return an error from a function call, wrap it.

handler.go
xlab: @soupdiver if the code will change to some extent that `http.NewRequest` will fail, we will need to panic or fatal here, because it's critical to initialize a new request. This will trigger a E2E error in CI one day.

If someone will be changing this particular function call, he may add error checking as well, that will match his scenario.

soupdiver: > he may add error checking as well, that will match his scenario.

People are lazy... if the error is omitted once, it will probably stay like that... because it was done like this before.

I already wasted hours of my life because I omitted errors "that should not happen" but for some reason still did.

If `http.NewRequest` returns an error for some reason, we end up here: https://github.com/cloudradar-monitoring/cagent/blob/4972b16a28779fb2a595458fe57ca233470f06ce/handler.go#L353 and post an error message. I think this is better than making an assumption that the case will never happen

cmd/cagent/winui.go
```diff
@@ -285,51 +267,120 @@ func windowsShowSettingsUI(ca *cagent.Cagent, installationMode bool) {
...
+		return
 	}
+	RunDialog(ui.MainWindow, ui.ErrorIcon,
+		"Error", "Please run this program with administrator priveleges.", func() {
+			os.Exit(1)
+		})
 }
 
-func checkAdmin() bool {
+func checkPermissions() bool {
```
nikita-vanyasin: well, this func definition looks strange now. It's unclear, what value will be returned. The old name was not good too, but a new one is even harder to understand until you look at the function body.

How about `checkIsAdminUser()` or `checkIsRunningWithElevatedPrivileges()` ?

xlab:
```
checkPermissions bool = true = has permissions
checkPermissions bool = false = doesn't have permissions
```

Running as admin is only a solution to permissions issue, not the prerequisite.
The function itself doesn't check if user is admin, it only checks disk access permissions.

> checkIsRunningWithElevatedPrivileges

This should be a comment for the function, not name :D

xlab: Researched about `\\\\.\\PHYSICALDRIVE0` notation. This works for me as regular user, without "Run as Administrator", so I assumed this checks disc access only. I will rename to `checkIsRunningWithElevatedPrivileges ` to avoid confusion.

PR #123, opened 2019-03-29
Monitor memory and CPU used by each process

nikita-vanyasin:
New fields added for each process:
```
cpu_avg_usage_percent
rss // Resident Set Size or Private Working Set Size (actual memory size used by process)
vms // Virtual Memory Size (for some systems like Windows it's very big by default, even if process does not use it)
memory_usage_percent
```

The cpu_avg_usage_percent calculated since last check, therefore it's available only after second check.

### DIFF
--- a/processes_notwindows.go
+++ b/processes_notwindows.go
@@ -15,9 +15,12 @@ import (
 	"strconv"
 	"strings"
 	"syscall"
+	"time"
 
+	"github.com/shirou/gopsutil/process"
 	log "github.com/sirupsen/logrus"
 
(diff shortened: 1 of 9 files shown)


processes_notwindows.go
```diff
@@ -29,12 +32,13 @@ type procStatus struct {
 }
 
 var dockerContainerIDRE = regexp.MustCompile(`(?m)/docker/([a-f0-9]*)$`)
+var monitoredProcessCache = make(map[int]*process.Process, 0)
```
golangcibot: S1019: should use make(map[int]*process.Process) instead (from `gosimple`)

PR #124, opened 2019-03-30
Report temperature sensor values

nikita-vanyasin:
See DEV-944

### DIFF
--- a/pkg/monitoring/sensors/temperature.go
+++ b/pkg/monitoring/sensors/temperature.go
@@ -1,48 +1,14 @@
 package sensors
 
-import (
-	"context"
-	"strings"
-	"time"
+import log "github.com/sirupsen/logrus"
 
-	"github.com/shirou/gopsutil/host"
-	log "github.com/sirupsen/logrus"
-)
-
-const readTimeout = time.Second * 10
 const unitCelsius = "centigrade"
 
 type TemperatureSensorInfo struct {
 	SensorName  string  `json:"sensor_name"`
 	Temperature float64 `json:"temperature"`
+	Critical    float64 `json:"critical"` // a temperature threshold set by sensor, driver or configuration
 	Unit        string  `json:"unit"`
 }
 
-func ReadTemperatureSensors() ([]*TemperatureSensorInfo, error) {
-	ctx, cancel := context.WithTimeout(context.Background(), readTimeout)
-	defer cancel()
-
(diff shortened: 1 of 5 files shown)


pkg/monitoring/sensors/temperature.go
```diff
@@ -0,0 +1,47 @@
...
+
+func ReadTemperatureSensors() ([]*TemperatureSensorInfo, error) {
+	ctx, cancel := context.WithTimeout(context.Background(), readTimeout)
+	defer cancel()
+
+	sensors, err := host.SensorsTemperaturesWithContext(ctx)
+	if err != nil {
+		errText := strings.ToLower(err.Error())
+		if strings.Contains(errText, "not implemented") {
+			logrus.Info("[SENSORS] reading temperature sensors is not implemented for this OS")
```
xlab: With logrus you need to use structure logging, so all context info is indexable for later use. Here `[SENSORS]` is a part of the message, however it should be a tag for the error. Must be like this:
```go
logrus.WithFields(logrus.Fields{
"pkg": "monitoring",
"module": "sensors",
}).Info("reading temperature sensors...")
```

Also, these fields can be set to a variable like this:

```go
moduleLog := logrus.WithFields(logrus.Fields{
"pkg": "monitoring",
"module": "sensors",
})

moduleLog.Info("reading temperature sensors is not implemented for this OS")
moduleLog.Info.WithError(err).Debugf("not supported by BIOS or driver required")
moduleLog.WithError(err).Error("failed to read temperature sensors")
```

nikita-vanyasin: Yeah, I know about this option. But in all parts of the project, this format is used: "[MODULE_NAME] some message here" (actually, in some modules the square brackets are missing).

I like the idea of introducing the log format that is simpler to parse, but we should agree on it before applying. Should we apply it? If yes, should we replace the old log.XXX() calls with the new format when making changes to file? I think having two different log formats for one file/module is not an option.

@soupdiver ?

soupdiver: See DEV-956
I created a ticket for the refactoring

nikita-vanyasin: I've updated the logging using package-level variable:
```
var logger = log.WithField("package", "sensors")
```
Also the bunch of changes regarding implementation: now it does not report useless values and reports a 'critical' threshold for each sensor (if available)

PR #127, opened 2019-04-09
Do not fail if there was any stderr output from ghw/pci library

nikita-vanyasin:
If some system file is missing, it's not an error. We will report it as 'info' message

### DIFF
--- a/pkg/hwinfo/hwinfo_other.go
+++ b/pkg/hwinfo/hwinfo_other.go
@@ -71,7 +71,8 @@ func listPCIDevices() ([]*pciDeviceInfo, error) {
 		return nil, errors.Wrap(ghwErr, "there were error while retrieving PCI information using ghw")
 	}
 	if len(stderrOutput) > 0 {
-		return nil, errors.Errorf("there were error output while retrieving PCI information using ghw: %s", stderrOutput)
+		// ghw only reports to stderr in case if system files are missing or there was failure while reading it
+		log.Infof("system is not supported: %s", stderrOutput)
 	}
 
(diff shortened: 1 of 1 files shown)


pkg/hwinfo/hwinfo_other.go
```diff
@@ -71,7 +71,8 @@ func listPCIDevices() ([]*pciDeviceInfo, error) {
 		return nil, errors.Wrap(ghwErr, "there were error while retrieving PCI information using ghw")
 	}
 	if len(stderrOutput) > 0 {
-		return nil, errors.Errorf("there were error output while retrieving PCI information using ghw: %s", stderrOutput)
+		// ghw only reports to stderr in case if system files are missing or there was failure while reading it
+		log.Infof("system is not supported: %s", stderrOutput)
```
soupdiver: I personally would use log level `Warn` I think... or was it discussed with @thorstenkramm to use `Info`?

nikita-vanyasin: Well, that's a good question. We use 'error' level in some packages in such cases, but some time it's even a 'debug' level.
I think 'warn' fits better. I also slightly improved the message.

thorstenkramm: In my opinion, `Info` is the best option. If the kernel does not support it, it's just information. A warning is something you should worry about and maybe take action. But there is nothing the user can do to solve the problem. (except re-compile the kernel) If we cannot gather this data, the user should not worry. Under the circumstances of the used kernel, it's normal behavior.

PR #128, opened 2019-04-09
Try catch dmidecode error accessing /dev/mem

nikita-vanyasin:
dmidecode can't read /dev/mem in some linux-systems (depending on kernel configuration options)

### DIFF
--- a/pkg/hwinfo/hwinfo_nonwindows.go
+++ b/pkg/hwinfo/hwinfo_nonwindows.go
@@ -6,7 +6,9 @@ import (
 	"bufio"
 	"bytes"
 	"fmt"
+	"io/ioutil"
 	"os/exec"
+	"strings"
 
 	"github.com/cloudradar-monitoring/dmidecode"
 	"github.com/pkg/errors"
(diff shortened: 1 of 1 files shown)


pkg/hwinfo/hwinfo_nonwindows.go
```diff
@@ -61,16 +63,27 @@ func retrieveInfoUsingDmiDecode() (map[string]interface{}, error) {
...
+	stderrBuffer := bytes.Buffer{}
+	cmd.Stderr = bufio.NewWriter(&stderrBuffer)
 
-	buf := bytes.Buffer{}
-	cmd.Stdout = bufio.NewWriter(&buf)
 	if err := cmd.Run(); err != nil {
+		stderrBytes, _ := ioutil.ReadAll(bufio.NewReader(&stderrBuffer))
+		stderr := string(stderrBytes)
+		if strings.Contains(stderr, "/dev/mem: Operation not permitted") {
+			log.Infof("[HWINFO] there was an error while executing '%s': %s\nProbably 'CONFIG_STRICT_DEVMEM' kernel configuration option is enabled. Please refer to kernel configuration manual.", dmidecodeCmd, stderr)
```
soupdiver: Is this `Info` log level and not rather `Warning` or `Error`?

nikita-vanyasin: Thorsten and I think `Info` is more appropriate because it's not a failure of cagent neither an error of system configuration. It's just info for the user that some data couldn't be acquired because the kernel configuration does not allow it.

Also, we use `info`  log level for similar cases in `hwinfo` package and others.

PR #133, opened 2019-04-16
Fix panic after unsuccessful attempt to gather process memory usage info

nikita-vanyasin:


soupdiver:
Please always(!) provide a description for a PR, no matter how small the PR is

PR #134, opened 2019-04-16
fix: run dmidecode without sudo on Darwin

troian:
on MacOS it provides result without being called as sudo
Call with sudo prompts for password and it fails due to
absence of `cagent-dmidecode` for MacOS package

### DIFF
--- a/pkg/hwinfo/hwinfo_nonwindows.go
+++ b/pkg/hwinfo/hwinfo_nonwindows.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"os/exec"
+	"runtime"
 	"strings"
 
 	"github.com/cloudradar-monitoring/dmidecode"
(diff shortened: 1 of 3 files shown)


pkg/hwinfo/hwinfo_nonwindows.go
```diff
@@ -63,9 +64,18 @@ func retrieveInfoUsingDmiDecode() (map[string]interface{}, error) {
...
 
-	// expecting 'sudo' package is installed and /etc/sudoers.d/cagent-dmidecode is present
-	const dmidecodeCmd = "sudo dmidecode"
-	cmd := exec.Command("/bin/sh", "-c", dmidecodeCmd)
+	var dmidecodeCmd []string
+
+	// run dmidecode as command argument to shell
+	dmidecodeCmd = append(dmidecodeCmd, "-c")
+
+	if runtime.GOOS != "darwin" {
```
nikita-vanyasin: I actually would prefer splitting implementation in compile-time where possible. But we probably should have some agreement with other developers for such cases.

I think, splitting implementation in compile time has a lot of advantages. We already have separate files for these cases (_darwin and _nonwindows), so we can create a function like `executeDmidecode` or `prepareDmidecodeCmd`

troian: IMO either case is ok. There was some discussion for that case some while ago and I remember we ended up with this approach. I might be wrong though...

soupdiver: I was thinking about the same thing.
I personally also prefer the compile-time way.
The function(s) could be written in a more abstract way and call a function like `smartctlCommand()` which returns  the string for the current OS. Only this helper would sit in `_linux _windows etc.` file

@xlab Any opinion?

troian: Made it to detect at compile time

PR #135, opened 2019-04-16
feat: S.M.A.R.T control

troian:
Parse and decode output provided by smartctl
- include prebuilt binary of `smartctl` for windows

### DIFF
--- a/cagent.go
+++ b/cagent.go
@@ -34,6 +34,8 @@ type Cagent struct {
 	vmstatLazyInit sync.Once
 	vmWatchers     map[string]types.Provider
 	hwInventory    sync.Once
+	initSMART      sync.Once
+	smartAvailable bool
 
 	rootCAs *x509.CertPool
 
@@ -47,6 +49,7 @@ func New(cfg *Config, cfgPath string, version string) *Cagent {
 		version:        version,
 		vmWatchers:     make(map[string]types.Provider),
 		dockerWatcher:  &docker.Watcher{},
+		smartAvailable: false,
 	}
 
 	if rootCertsPath != "" {
--- a/handler.go
+++ b/handler.go
@@ -6,6 +6,7 @@ import (
 	"context"
 	"crypto/tls"
 	"encoding/json"
+	"fmt"
 	"io"
 	"io/ioutil"
 	"net/http"
@@ -25,6 +26,7 @@ import (
 	"github.com/cloudradar-monitoring/cagent/pkg/monitoring/services"
 	"github.com/cloudradar-monitoring/cagent/pkg/monitoring/vmstat"
 	vmstatTypes "github.com/cloudradar-monitoring/cagent/pkg/monitoring/vmstat/types"
+	"github.com/cloudradar-monitoring/cagent/pkg/smart"
 )
 
 func (ca *Cagent) initHubClientOnce() {
@@ -269,10 +271,14 @@ func (ca *Cagent) GetAllMeasurements() (common.MeasurementsMap, error) {
 
 	ca.hwInventory.Do(func() {
 		hwInfo, err := hwinfo.Inventory()
+
 		if err != nil {
 			errs = append(errs, err.Error())
 		}
-		measurements = measurements.AddInnerWithPrefix("hw.inventory", hwInfo)
+
+		if hwInfo != nil {
+			measurements = measurements.AddInnerWithPrefix("hw.inventory", hwInfo)
+		}
 	})
 
 	if runtime.GOOS == "linux" {
@@ -333,13 +339,26 @@ func (ca *Cagent) GetAllMeasurements() (common.MeasurementsMap, error) {
 
 	if len(errs) == 0 {
 		measurements["cagent.success"] = 1
-
-		return measurements, nil
+	} else {
+		measurements["cagent.success"] = 0
 	}
 
-	measurements["cagent.success"] = 0
+	// measurements fetched below should not affect cagent.success
+	smartMeas, smartErrs := ca.getSMARTMeasurements()
 
-	return measurements, errors.New(strings.Join(errs, "; "))
+	if len(smartMeas) > 0 {
+		measurements = measurements.AddInnerWithPrefix("smartmon", smartMeas)
+	}
+
+	for _, e := range smartErrs {
+		errs = append(errs, e.Error())
+	}
+
+	if len(errs) != 0 {
+		return measurements, errors.New(strings.Join(errs, "; "))
+	}
+
+	return measurements, nil
 }
 
 func (ca *Cagent) ReportMeasurements(measurements common.MeasurementsMap, outputFile *os.File) error {
@@ -415,4 +434,28 @@ func (ca *Cagent) getVMStatMeasurements(f func(string, common.MeasurementsMap, e
 		res, err := p.GetMeasurements()
 		f(name, res, err)
 	}
+}
+
+func (ca *Cagent) getSMARTMeasurements() (common.MeasurementsMap, []error) {
+	// measurements fetched below should not affect cagent.success
+	ca.initSMART.Do(func() {
+		if ca.Config.SMARTMonitoring {
+			if err := smart.DetectTools(); err == nil {
+				ca.smartAvailable = true
+			} else {
+				ca.smartAvailable = false
+				log.Error(err.Error())
+			}
+		} else {
+			ca.smartAvailable = false
+		}
+	})
+
+	if ca.smartAvailable {
+		res, errs := smart.Parse()
+
+		return res, errs
+	}
+
+	return common.MeasurementsMap{}, []error{fmt.Errorf("smartmon: not available")}
 }
--- a/handler.go
+++ b/handler.go
@@ -271,7 +271,6 @@ func (ca *Cagent) GetAllMeasurements() (common.MeasurementsMap, error) {
 
 	ca.hwInventory.Do(func() {
 		hwInfo, err := hwinfo.Inventory()
-
 		if err != nil {
 			errs = append(errs, err.Error())
 		}
--- a/pkg/smart/parse.go
+++ b/pkg/smart/parse.go
@@ -1,0 +1,333 @@
+package smart
+
+import (
+	"bufio"
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"os/exec"
+	"regexp"
+	"strconv"
+	"strings"
+	"syscall"
+
+	"github.com/pkg/errors"
+
+	"github.com/cloudradar-monitoring/cagent/pkg/common"
+)
+
+type ataSMARTAttributes struct {
+	Table []struct {
+		ID         int    `json:"id"`
+		Name       string `json:"name"`
+		Value      int    `json:"value"`
+		Worst      int    `json:"worst"`
+		Thresh     int    `json:"thresh"`
+		WhenFailed string `json:"when_failed"`
+		Raw        struct {
+			Value  int    `json:"value"`
+			String string `json:"string"`
+		} `json:"raw"`
+	} `json:"table"`
+}
+
+type nvmeSmartHealthInformationLog struct {
+	CriticalWarning         int   `json:"critical_warning"`
+	Temperature             int   `json:"temperature"`
+	AvailableSpare          int   `json:"available_spare"`
+	AvailableSpareThreshold int   `json:"available_spare_threshold"`
+	PercentageUsed          int   `json:"percentage_used"`
+	DataUnitsRead           int64 `json:"data_units_read"`
+	DataUnitsWritten        int64 `json:"data_units_written"`
+	HostReads               int64 `json:"host_reads"`
+	HostWrites              int64 `json:"host_writes"`
+	ControllerBusyTime      int64 `json:"controller_busy_time"`
+	PowerCycles             int   `json:"power_cycles"`
+	PowerOnHours            int   `json:"power_on_hours"`
+	UnsafeShutdowns         int   `json:"unsafe_shutdowns"`
+	MediaErrors             int   `json:"media_errors"`
+	NumErrLogEntries        int   `json:"num_err_log_entries"`
+}
+
+type parseResult struct {
+	Smartctl struct {
+		Status   int `json:"exit_status"`
+		Messages []struct {
+			String string `json:"string"`
+		}
+	} `json:"smartctl"`
+	Device struct {
+		Name     string `json:"name"`
+		InfoName string `json:"info_name"`
+		Type     string `json:"type"`
+		Protocol string `json:"protocol"`
+	} `json:"device"`
+
+	ModelName          string  `json:"model_name"`
+	SerialNumber       string  `json:"serial_number"`
+	ModelFamily        *string `json:"model_family"`
+	FirmwareVersion    string  `json:"firmware_version"`
+	InSmartctlDatabase bool    `json:"in_smartctl_database"`
+
+	SmartStatus *struct {
+		Passed bool `json:"passed"`
+	} `json:"smart_status"`
+
+	Temperature *struct {
+		Current float32 `json:"current"`
+	} `json:"temperature"`
+
+	PowerCycleCount *int `json:"power_cycle_count"`
+	PowerOnTime     *struct {
+		Hours int `json:"hours"`
+	} `json:"power_on_time"`
+
+	RotationRate      *string `json:"rotation_rate,omitempty"`
+	LogicalBlockSize  *int    `json:"logical_block_size,omitempty"`
+	PhysicalBlockSize *int    `json:"physical_block_size,omitempty"`
+
+	UserCapacity *struct {
+		Blocks int   `json:"blocks"`
+		Bytes  int64 `json:"bytes"`
+	} `json:"user_capacity"`
+
+	ATAVersion *struct {
+		String     string `json:"string"`
+		MajorValue int    `json:"major_value"`
+		MinorValue int    `json:"minor_value"`
+	} `json:"ata_version"`
+
+	SATAVersion *struct {
+		String string `json:"string"`
+		Value  int    `json:"value"`
+	} `json:"sata_version"`
+
+	InterfaceSpeed *struct {
+		Max struct {
+			SataValue      int    `json:"sata_value"`
+			String         string `json:"string"`
+			UnitsPerSecond int64  `json:"units_per_second"`
+			BitsPerUnit    int64  `json:"bits_per_unit"`
+		} `json:"max"`
+	} `json:"interface_speed"`
+
+	NVMESmartHealthInformationLog *nvmeSmartHealthInformationLog `json:"nvme_smart_health_information_log"`
+
+	ATASmartData *struct {
+		SelfTest struct {
+			String string `json:"string"`
+			Passed bool   `json:"passed"`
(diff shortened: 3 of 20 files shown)


cagent.go
```diff
@@ -34,6 +34,8 @@ type Cagent struct {
 	vmstatLazyInit sync.Once
 	vmWatchers     map[string]types.Provider
 	hwInventory    sync.Once
+	smartInit      sync.Once
```
soupdiver: I suggest to change `smart` => `SMART` as done in config.go

handler.go
```diff
@@ -332,11 +333,36 @@ func (ca *Cagent) GetAllMeasurements() (MeasurementsMap, error) {
...
+		res, err := smart.Parse()
+		measurements = measurements.AddInnerWithPrefix("smartmon", res)
+
+		for _, e := range err {
+			errs = append(errs, e.Error())
+		}
 	}
 
-	measurements["cagent.success"] = 0
+	if len(errs) == 0 {
```
soupdiver: I would check against `len(errs) != 0` and in case return `measurements, errors.New(strings.Join(errs, "; "))`

It's just a nuance, but usually we check for error/special cases and then return. The final return should be the success case.

pkg/smart/checkRequiredTools_darwin.go
```diff
@@ -0,0 +1,27 @@
+// +build darwin
```
soupdiver: Please use snake-case instead camel-case for filename

pkg/smart/checkRequiredTools_linux.go
```diff
@@ -0,0 +1,22 @@
+// +build !darwin,!windows
```
soupdiver: Filename

pkg/smart/detectDisks_darwin.go
```diff
@@ -0,0 +1,25 @@
+// +build darwin
```
soupdiver: Filename

pkg/smart/checkRequiredTools_linux.go
```diff
@@ -0,0 +1,22 @@
...
+	"os/exec"
+)
+
+func checkTools() (string, error) {
+	cmd := exec.Command("/bin/sh", "-c", "smartctl -h")
+	buf := &bytes.Buffer{}
+	cmd.Stdout = bufio.NewWriter(buf)
+
+	if err := cmd.Run(); err != nil {
+		return "", fmt.Errorf("smart: smartctl is not installed")
```
soupdiver: Why not `ErrUnderlyingToolNotFound ` here?

pkg/smart/checkRequiredTools_darwin.go
```diff
@@ -0,0 +1,27 @@
...
+	"os/exec"
+)
+
+func checkTools() (string, error) {
+	cmd := exec.Command("/bin/sh", "-c", "smartctl -h")
+	buf := &bytes.Buffer{}
+	cmd.Stdout = bufio.NewWriter(buf)
+
+	if err := cmd.Run(); err != nil {
+		return "", fmt.Errorf("smart: smartctl is not installed")
```
soupdiver: Why not `ErrUnderlyingToolNotFound` here?

pkg/smart/detectDisks_darwin.go
```diff
@@ -0,0 +1,25 @@
...
+)
+
+func detectDisks() (*bytes.Buffer, error) {
+	cmd := exec.Command("/bin/sh", "-c", `diskutil list | grep "^/dev/" | grep -v synthesized | grep -v external | grep -v "disk image"`)
+
+	buf := &bytes.Buffer{}
+	cmd.Stdout = bufio.NewWriter(buf)
+
+	if err := cmd.Run(); err != nil {
+		log.Errorf("smart: execute diskutil: %s", err.Error())
```
soupdiver: Why do we log _and_ return the error here?

Applies for further occasions as well

pkg/smart/detectDisks_windows.go
```diff
@@ -0,0 +1,25 @@
+// +build windows
```
soupdiver: Filename

pkg/smart/parse.go
```diff
@@ -0,0 +1,329 @@
...
+	"os/exec"
+	"regexp"
+	"strconv"
+	"strings"
+	"syscall"
+
+	"github.com/pkg/errors"
+)
+
+type ataSmartAttributes struct {
```
soupdiver: Use `ATASMARTData`?

pkg/smart/parse.go
```diff
@@ -0,0 +1,329 @@
...
+	HostWrites              int64 `json:"host_writes"`
+	ControllerBusyTime      int64 `json:"controller_busy_time"`
+	PowerCycles             int   `json:"power_cycles"`
+	PowerOnHours            int   `json:"power_on_hours"`
+	UnsafeShutdowns         int   `json:"unsafe_shutdowns"`
+	MediaErrors             int   `json:"media_errors"`
+	NumErrLogEntries        int   `json:"num_err_log_entries"`
+}
+
+type parseResult struct {
```
soupdiver: Is it the beast approach to define all these as anonymous types?

pkg/smart/parse.go
```diff
@@ -0,0 +1,329 @@
...
+var smartctlVersionRegexp *regexp.Regexp
+
+func init() {
+	smartctlVersionRegexp = regexp.MustCompile(`^smartctl\s(\d.\d)\s(\w|\W)+$`)
+}
+
+func DetectTools() error {
+	buildStr, err := checkTools()
+	if err != nil {
+		return err
```
soupdiver: Wrap the error?

types.go
```diff
@@ -14,7 +14,7 @@ func (mm MeasurementsMap) AddWithPrefix(prefix string, m MeasurementsMap) Measur
 	return mm
 }
 
-func (mm MeasurementsMap) AddInnerWithPrefix(prefix string, m MeasurementsMap) MeasurementsMap {
+func (mm MeasurementsMap) AddInnerWithPrefix(prefix string, m interface{}) MeasurementsMap {
```
soupdiver: Why do we change the type to `interface{}`?

handler.go
```diff
@@ -332,11 +333,36 @@ func (ca *Cagent) GetAllMeasurements() (MeasurementsMap, error) {
 
 	if len(errs) == 0 {
 		measurements["cagent.success"] = 1
+	} else {
+		measurements["cagent.success"] = 0
+	}
 
-		return measurements, nil
+	// measurements fetched below should not affect cagent.success
+	ca.smartInit.Do(func() {
```
nikita-vanyasin: In all other checks, they were checking for feature availability in its implementation. (And returning nil result in case if some package is not installed or feature disabled by cagent cofiguration). I think this check and smartmon parsing should be moved to some function that will abstract this checks. Currently, the `GetAllMeasurements` function became even more inconsistent.

pkg/smart/parse.go
nikita-vanyasin: The correct naming is `ataSMARTAttributes` or `ataSMARTData`

pkg/smart/parse.go
```diff
@@ -0,0 +1,329 @@
...
+		Flags struct {
+			ReminderScanEnabled bool `json:"reminder_scan_enabled"`
+		} `json:"flags"`
+		PowerUpScanResumeMinutes int `json:"power_up_scan_resume_minutes"`
+	} `json:"ata_smart_selective_self_test_log"`
+}
+
+var smartctlVersionRegexp *regexp.Regexp
+
+func init() {
```
nikita-vanyasin: Not sure why we need this init() function.
The regexp variable can be defined directly in its definition. See example from `hwinfo_other.go:23`
`var lsusbLineRegexp = regexp.MustCompile(`[0-9|a-z|A-Z|.|/|-|:|\[|\]|_|+| ]+`)`

pkg/smart/parseDisk.go
```diff
@@ -0,0 +1,40 @@
+package smart
+
+import (
+	"bytes"
+	"regexp"
+)
+
+var diskScanRegexp *regexp.Regexp
+
+func init() {
```
nikita-vanyasin: same comment about regexp variable initialization (see above)

pkg/smart/parseDisk.go
```diff
@@ -0,0 +1,40 @@
...
+func parseDisks(buf *bytes.Buffer) ([]string, error) {
+	var disks []string
+
+	for {
+		s, err := buf.ReadString('\n')
+		if s != "" {
+			disks = append(disks, s)
+		}
+
+		if err != nil {
```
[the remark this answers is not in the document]
nikita-vanyasin: Probably this check should also be altered: 
err can have  EOF value here, which will end the loop. But also it can have some error and we probably should handle it..

pkg/smart/parse.go
```diff
@@ -0,0 +1,329 @@
...
+	buildStr, err := checkTools()
+	if err != nil {
+		return err
+	}
+
+	return smartctlIsSupportedVersion(buildStr)
+}
+
+// Parse detect hardware disks and parse their S.M.A.R.T
+func Parse() (interface{}, []error) {
```
nikita-vanyasin: Returning a strict type instead of interface{} (which is actually map[string]) can make the code more readable and consistent with other packages: almost all of them use struct definitions with `json` marshaling settings for returning result.

nikita-vanyasin:
> I wonder if it's really the best way to store 3rd party binaries inside the repository.
> I can imagine more tools might be added and then it gets out of hand.
> Maybe we should think about another way of providing these 3rd party tools. @thorstenkramm

I actually curious why it was decided to include the `smartmon`, but not to include `dmidecode` binary. We need some consistency here :)

troian:
@nikita-vanyasin `dmidecode` is linux specific and can be marked as dependency to cagent deb or rpm package. We do same for smartmontools on Linux platforms.

On Windows there is no way to install some 3rd party packet. Thats why smartctl.exe included into distribution

cagent.go
troian: Resolved?

handler.go

pkg/smart/checkRequiredTools_linux.go
troian: done

pkg/smart/checkRequiredTools_darwin.go
troian: done

pkg/smart/detectDisks_darwin.go
troian: done

pkg/smart/smartctlPrepare_notwindows.go
```diff
@@ -0,0 +1,18 @@
+// +build !windows
```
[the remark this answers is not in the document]
troian: done

pkg/smart/smartctlVersion_test.go
```diff
@@ -0,0 +1,30 @@
+package smart
```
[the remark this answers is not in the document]
troian: done

handler.go
```diff
@@ -332,11 +333,36 @@ func (ca *Cagent) GetAllMeasurements() (MeasurementsMap, error) {
 
 	if len(errs) == 0 {
 		measurements["cagent.success"] = 1
+	} else {
+		measurements["cagent.success"] = 0
+	}
 
-		return measurements, nil
+	// measurements fetched below should not affect cagent.success
```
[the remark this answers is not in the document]
troian: That's agreed rule of `cagent`. If any 3rd-party measurement tool fails it should not affect `cagent.success`. There were couple discussions in slack

types.go
troian: Using `MeasurementsMap` causes cyclic dependency issue.

handler.go
```diff
@@ -267,11 +268,11 @@ func (ca *Cagent) GetAllMeasurements() (MeasurementsMap, error) {
 	})
 
 	ca.hwInventory.Do(func() {
-		hwInfo, err := hwinfo.Inventory()
-		if err != nil {
+		if hwInfo, err := hwinfo.Inventory(); err != nil {
 			errs = append(errs, err.Error())
+		} else {
```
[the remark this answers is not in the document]
troian: Reverted. It came in from some old merge

handler.go
troian: I would leave as is till common approach for `GetAllMeasurements` found

pkg/smart/parse.go
troian: ok

pkg/smart/parseDisk.go
troian: done

pkg/smart/parse.go
troian: Please look to answer above related to `AddInnerWithPrefix`

handler.go
soupdiver: @troian Why do you mark this as resolved if no change happened?

cagent.go
troian: Or initSMART should work as well

soupdiver: I would prefer `initSMART` then

handler.go
soupdiver: I agree with @nikita-vanyasin 
The function is already quite big but so far it did not do checking if things are available or not. If something is not available or caused an error, the underlying function returns an empty struct, not `nil`, and the errors. I think we should continue like this

pkg/smart/parse.go
soupdiver: This concern was raised already in the past, that we should do a proper re-factoring to be able to use `MeasurementMap` instead `interface{}`.
Guess for now we can't do much :/

troian: https://github.com/cloudradar-monitoring/cagent/pull/136

soupdiver:
Should we merge #136 first and then update this PR?
Also remove the binary and we integrate this into the MSI during CI?
@troian

troian:
@soupdiver 
yes for both. I'll force-push without binary when #136 is merged

handler.go

pkg/smart/parse.go
troian: done

cagent.go

pkg/smart/parse.go
```diff
@@ -0,0 +1,329 @@
...
+	}
+
+	result, parseErrors := smartCtlParse(jsonOutput)
+
+	return result, append(errs, parseErrors...)
+}
+
+func smartCtlRun(disks []string) ([]string, error) {
+	var result []string
+
```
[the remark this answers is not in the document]
troian: done

pkg/smart/parse.go
```diff
@@ -0,0 +1,329 @@
...
+		if err != nil {
+			errs = append(errs, err)
+		} else {
+			parsedDisks = append(parsedDisks, res)
+		}
+	}
+
+	var marshaledDisks []interface{}
+
+	for _, d := range parsedDisks {
```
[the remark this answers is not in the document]
troian: `d` is short to `disk`

pkg/smart/parse.go
```diff
@@ -0,0 +1,329 @@
...
+	}
+
+	return result, nil
+}
+
+func smartCtlParse(raw []string) (interface{}, []error) {
+	var parsedDisks []*parseResult
+
+	var errs []error
+	for _, r := range raw {
```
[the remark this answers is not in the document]
troian: `r` is short to `raw`

pkg/smart/parse.go
troian: checkTools returns error with `smart:` prefix

pkg/smart/parse.go
```diff
@@ -0,0 +1,329 @@
...
+
+		marshaledDisks = append(marshaledDisks, output)
+	}
+
+	return marshaledDisks, errs
+}
+
+func smartctlIsSupportedVersion(buildStr string) error {
+	if !smartctlVersionRegexp.Match([]byte(buildStr)) {
+		return errors.New("smart: couldn't detect smartctl version")
```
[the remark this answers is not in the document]
troian: It appears only once

pkg/smart/parse.go
```diff
@@ -0,0 +1,329 @@
...
+	}
+
+	ver := smartctlVersionRegexp.FindAllStringSubmatch(buildStr, -1)
+
+	tok := strings.Split(ver[0][1], ".")
```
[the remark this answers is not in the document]
troian: It is validated by `smartctlVersionRegexp.Match` before extracting tokens

pkg/smart/parse.go
```diff
@@ -0,0 +1,329 @@
...
+	if err != nil {
+		return fmt.Errorf("smart: parse smartctl version: %s", err)
+	}
+
+	minor, err = strconv.Atoi(tok[1])
+	if err != nil {
+		return fmt.Errorf("smart: parse smartctl version: %s", err)
+	}
+
+	if (major > 7) || ((major == 7) && (minor >= 0)) {
```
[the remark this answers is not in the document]
troian: done

pkg/smart/parse.go
```diff
@@ -0,0 +1,329 @@
...
+	var result []string
+
+	var errStr string
+
+	for _, disk := range disks {
+		cmd := smartctlPrepare(disk)
+		buf := &bytes.Buffer{}
+		cmd.Stdout = bufio.NewWriter(buf)
+
+		if err := cmd.Run(); err != nil {
```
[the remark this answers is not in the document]
troian: done

pkg/smart/parse.go
troian: These fields are internal use only to parse `json` output from `smartctl`.

pkg/smart/parse.go
```diff
@@ -0,0 +1,338 @@
...
+	"strings"
+	"syscall"
+
+	"github.com/pkg/errors"
+
+	"github.com/cloudradar-monitoring/cagent/pkg/common"
+)
+
+const (
+	smartCtlMinMajor = 7
```
golangcibot: `smartCtlMinMajor` is unused (from `deadcode`)

pkg/smart/parse.go
```diff
@@ -0,0 +1,338 @@
...
+	smartCtlMinMinor = 1
```
golangcibot: `smartCtlMinMinor` is unused (from `deadcode`)

pkg/smart/parse.go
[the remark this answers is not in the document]
soupdiver: I know 😅 
My point was to use `disk` as a variable name instead just `d`. 
Makes the code more readable

pkg/smart/parse.go
soupdiver: Wrapping the error is not for the prefix. It preserves the stack trace as well

pkg/smart/parse.go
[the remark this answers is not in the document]
soupdiver: hmmm weird... maybe it was updated or I'm wrong 🤔

handler.go
```diff
@@ -269,10 +271,14 @@ func (ca *Cagent) GetAllMeasurements() (common.MeasurementsMap, error) {
 
 	ca.hwInventory.Do(func() {
 		hwInfo, err := hwinfo.Inventory()
+
```

PR #136, opened 2019-04-23
feat: allocate MeasurementsMap to package

troian:
Allows using `MeasurementsMap` in all packages returning measurements without cyclic dependency issue


mem.go
```diff
@@ -10,12 +10,14 @@ import (
...
+
+	"github.com/cloudradar-monitoring/cagent/types"
 )
 
 const memGetTimeout = time.Second * 10
 
-func (ca *Cagent) MemResults() (MeasurementsMap, *mem.VirtualMemoryStat, error) {
-	results := MeasurementsMap{}
+func (ca *Cagent) MemResults() (types.MeasurementsMap, *mem.VirtualMemoryStat, error) {
+	results := types.MeasurementsMap{}
```
golangcibot: SA4006: this value of `results` is never used (from `staticcheck`)

windows.go
```diff
@@ -5,6 +5,7 @@ package cagent
 import (
 	"time"
 
+	"github.com/cloudradar-monitoring/cagent/types"
```
soupdiver: external imports come first followed by a new line and then our internal imports

nikita-vanyasin:
> I wonder if we should move all things from `types.go` to `types/types.go`
> Then we only have one place where we define our types.
> Opinions?

I think it's a good idea to move all common functions from types.go
I also think it's better to move common definitions to existing package `common` and not to introduce a new package `types`.

windows.go
troian: done

troian:
> > I wonder if we should move all things from `types.go` to `types/types.go`
> > Then we only have one place where we define our types.
> > Opinions?
> 
> I think it's a good idea to move all common functions from types.go
> I also think it's better to move common definitions to existing package `common` and not to introduce a new package `types`.

Sounds reasonable, but might be package common should go to top level
@soupdiver ?

soupdiver:
> Sounds reasonable, but might be package common should go to top level

It should be `pkg/common`

troian:
All in pkg/common

fs.go
```diff
@@ -7,6 +7,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/cloudradar-monitoring/cagent/pkg/common"
```
golangcibot: File is not `goimports`-ed (from `goimports`)

mem.go
```diff
@@ -8,18 +8,18 @@ import (
 	"runtime"
 	"time"
 
+	"github.com/cloudradar-monitoring/cagent/pkg/common"
```
golangcibot: File is not `goimports`-ed (from `goimports`)

network.go
```diff
@@ -8,6 +8,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/cloudradar-monitoring/cagent/pkg/common"
```
golangcibot: File is not `goimports`-ed (from `goimports`)

fs.go
troian: done

mem.go
troian: done

network.go
troian: done

ports.go
```diff
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"syscall"
 
+	"github.com/cloudradar-monitoring/cagent/pkg/common"
```
golangcibot: File is not `goimports`-ed (from `goimports`)

processes.go
```diff
@@ -3,6 +3,7 @@ package cagent
 import (
 	"runtime"
 
+	"github.com/cloudradar-monitoring/cagent/pkg/common"
```
golangcibot: File is not `goimports`-ed (from `goimports`)

raid.go
```diff
@@ -6,6 +6,7 @@ import (
 	"regexp"
 	"strings"
 
+	"github.com/cloudradar-monitoring/cagent/pkg/common"
```
golangcibot: File is not `goimports`-ed (from `goimports`)

swap.go
```diff
@@ -6,14 +6,15 @@ import (
 	"strings"
 	"time"
 
+	"github.com/cloudradar-monitoring/cagent/pkg/common"
```
golangcibot: File is not `goimports`-ed (from `goimports`)

system.go
```diff
@@ -11,6 +11,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/cloudradar-monitoring/cagent/pkg/common"
```
golangcibot: File is not `goimports`-ed (from `goimports`)

soupdiver:
I renamed the file `measurements.go` => `types.go`

processes.go
troian: done

ports.go
troian: done

swap.go
troian: done

system.go
troian: done

PR #143, opened 2019-05-02
feat: add option to pretty print hub measurements to file

troian:
Add config option to pretty print hub measurements to file
Useful for debug purpose

### DIFF
--- a/config.go
+++ b/config.go
@@ -38,6 +38,9 @@ type MinValuableConfig struct {
 	HubPassword string   `toml:"hub_password" commented:"true"`
 }
 
+type ConfigsLogs struct {
+	HubFile string `toml:"hub_file,omitempty" comment:"log hub objects send to the hub"`
+}
 type Config struct {
 	Interval float64 `toml:"interval" comment:"interval to push metrics to the HUB"`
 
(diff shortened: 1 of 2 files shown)


config.go
```diff
@@ -38,6 +38,9 @@ type MinValuableConfig struct {
 	HubPassword string   `toml:"hub_password" commented:"true"`
 }
 
+type ConfigsLogs struct {
+	HubFile string `toml:"hub_file,omitempty" comment:"log hub objects send to the hub"`
+}
```
nikita-vanyasin: Empty line missing

config.go
```diff
@@ -38,6 +38,9 @@ type MinValuableConfig struct {
 	HubPassword string   `toml:"hub_password" commented:"true"`
 }
 
+type ConfigsLogs struct {
```
nikita-vanyasin: Not sure why both words in the type name are plural. I think it's better to name like `LogFilesConfig`

config.go
troian: done

config.go
troian: done

PR #147, opened 2019-05-09
fix: detect smartctl executable

troian:
Add config param with smartctl executable
Detect location of smartctl
Fix smartctl execution on Win

### DIFF
--- a/pkg/smart/check_required_tools_darwin.go
+++ b/pkg/smart/check_required_tools_darwin.go
@@ -7,21 +7,36 @@ import (
 	"bytes"
 	"fmt"
 	"os/exec"
+	"strings"
+
+	"github.com/pkg/errors"
 )
 
-func checkTools() (string, error) {
-	cmd := exec.Command("/bin/sh", "-c", "smartctl -h")
-	buf := &bytes.Buffer{}
-	cmd.Stdout = bufio.NewWriter(buf)
+const defaultSmartctlPath = "smartctl"
+
+func checkTools(smartctl string) (string, string, error) {
+	cmd := exec.Command("/bin/sh", "-c", fmt.Sprintf("which %s", smartctl))
+	smartctlPathBuf := &bytes.Buffer{}
+	cmd.Stdout = bufio.NewWriter(smartctlPathBuf)
 
 	if err := cmd.Run(); err != nil {
-		return "", fmt.Errorf("smart: smartctl is not installed")
+		return "", "", errors.Wrapf(ErrSmartctlNotFound, "\"%s\"", smartctl)
+	}
+
+	smartctl = strings.TrimRight(smartctlPathBuf.String(), "\n")
+
+	cmd = exec.Command("/bin/sh", "-c", fmt.Sprintf("command %s -h", smartctl))
+	buildString := &bytes.Buffer{}
+	cmd.Stdout = bufio.NewWriter(buildString)
+	if err := cmd.Run(); err != nil {
+		return "", "", errors.Wrap(err, "smart: cannot get smartctl version string")
 	}
 
 	cmd = exec.Command("/bin/sh", "-c", "command -v diskutil")
+	cmd.Stdout = nil
 	if err := cmd.Run(); err != nil {
-		return "", fmt.Errorf("smart: diskutil is not installed")
+		return "", "", fmt.Errorf("smart: diskutil is not installed")
 	}
 
-	return buf.String(), nil
+	return buildString.String(), smartctl, nil
 }
--- a/pkg/smart/check_required_tools_linux.go
+++ b/pkg/smart/check_required_tools_linux.go
@@ -7,16 +7,28 @@ import (
 	"bytes"
 	"fmt"
 	"os/exec"
+	"strings"
+
+	"github.com/pkg/errors"
 )
 
-func checkTools() (string, error) {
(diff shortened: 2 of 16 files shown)


pkg/smart/check_required_tools_darwin.go
```diff
@@ -7,21 +7,36 @@ import (
...
 )
 
-func checkTools() (string, error) {
-	cmd := exec.Command("/bin/sh", "-c", "smartctl -h")
-	buf := &bytes.Buffer{}
-	cmd.Stdout = bufio.NewWriter(buf)
+const defaultSmartctlPath = "smartctl"
+
+func checkTools(smartctl string) (string, string, error) {
+	cmd := exec.Command("/bin/sh", "-c", fmt.Sprintf("which %s", smartctl))
```
martinlindhe: This code for getting path of smartctl seems excessive.

What about something like
```go
smartctlPathBuf, err := exec.Command("which", smartctl).Output()
if err != nil {
    return "", "", errors.Wrapf(ErrSmartctlNotFound, "\"%s\"", smartctl)
}
smartctl = strings.TrimRight(string(smartctlPathBuf), "\n")
```

pkg/smart/check_required_tools_linux.go
```diff
@@ -7,16 +7,28 @@ import (
...
+	"strings"
+
+	"github.com/pkg/errors"
 )
 
-func checkTools() (string, error) {
-	cmd := exec.Command("/bin/sh", "-c", "smartctl -h")
+const defaultSmartctlPath = "smartctl"
+
+func checkTools(smartctl string) (string, string, error) {
```
martinlindhe: Should be refactored and reused in _darwin.go

troian: I think linux and darwin should be separate

pkg/smart/check_required_tools_darwin.go
troian: Doubt it will work, where needs to run in cmd I think

martinlindhe: There's really no reason to spawn a shell just to run a command.

I did this test locally while coming up with the comment. It works

```go
package main

import (
    "fmt"
    "os/exec"
)

func main() {
    out, _ := exec.Command("which", "bash").Output()
    fmt.Println(string(out))
}
```

pkg/smart/check_required_tools_linux.go
martinlindhe: Ok.

martinlindhe: I'm not saying you should drop this file. I mean it would be preferred to refactor the common code instead of repeating it.

pkg/smart/check_required_tools_darwin.go
martinlindhe: Let's consider this in a upcoming PR instead

pkg/smart/check_required_tools_linux.go
martinlindhe: Let's consider this in a upcoming PR instead

pkg/smart/parse.go
```diff
@@ -309,15 +325,30 @@ func parseBase(output map[string]interface{}, d *parseResult) string {
...
+		}
+	case int:
+		if rt == 0 {
+			output["type_of"] = "SSD"
+		} else {
+			output["type_of"] = "HDD"
+			output["rotation_rate"] = rt
+		}
+	default:
+		logrus.Warnf("smart: parse type \"%s\" of rotation rate field of disk \"d.Device.Name\"", reflect.TypeOf(d.RotationRate).String())
```
nikita-vanyasin: Please use alias for `logrus`

pkg/smart/parse.go
```diff
@@ -245,36 +257,40 @@ func smartCtlParse(raw []string) (common.MeasurementsMap, []error) {
 	return marshaledDisks, errs
 }
 
-func smartctlIsSupportedVersion(buildStr string) error {
+func smartctlIsSupportedVersion(buildStr string) (string, error) {
```
nikita-vanyasin: the function name is kind of confusing.
It's actually confusing because it makes two things:
1) parsing version string
2) checking for compatibility
Splitting the function will simplify the code

pkg/smart/parse.go
```diff
@@ -245,36 +257,40 @@ func smartCtlParse(raw []string) (common.MeasurementsMap, []error) {
...
-		return fmt.Errorf("smart: parse smartctl version: %s", err)
+	if major, err = strconv.Atoi(tok[0]); err != nil {
+		return "", fmt.Errorf("smart: parse smartctl version: %s", err)
+	}
+
+	if minor, err = strconv.Atoi(tok[1]); err != nil {
+		return "", fmt.Errorf("smart: parse smartctl version: %s", err)
 	}
 
 	if major >= 7 {
```
nikita-vanyasin: Probably should be moved to `const` value.

pkg/smart/parse.go
troian: done

pkg/smart/parse.go
troian: done

pkg/smart/parse.go
troian: done

PR #149, opened 2019-05-15
feat: add fs config to exclude mount binds

troian:


fs.go
```diff
@@ -29,13 +30,19 @@ func (ca *Cagent) FSWatcher() *FSWatcher {
...
 		ExcludedPathCache: map[string]bool{},
 		cagent:            ca,
 	}
+
 	for _, t := range ca.Config.FSTypeInclude {
 		ca.fsWatcher.AllowedTypes[strings.ToLower(t)] = struct{}{}
 	}
 
+	for _, t := range ca.Config.FSPathExclude {
+		ca.fsWatcher.ExcludePath[strings.ToLower(t)] = struct{}{}
```
martinlindhe: I think path should not be lowercased here. Case matters on Linux

fs.go
```diff
@@ -49,25 +56,42 @@ func (fw *FSWatcher) Results() (common.MeasurementsMap, error) {
...
+				}
+			}
+		}
+
+		if pathExcluded {
+			continue
+		}
+
+		cacheExists := false
+		if pathExcluded, cacheExists = fw.ExcludedPathCache[strings.ToLower(partition.Mountpoint)]; cacheExists {
```
martinlindhe: Again, not sure about lowercasing paths

fs.go
troian: done

fs.go
troian: done

fs.go
```diff
@@ -36,11 +36,11 @@ func (ca *Cagent) FSWatcher() *FSWatcher {
 	}
 
 	for _, t := range ca.Config.FSTypeInclude {
-		ca.fsWatcher.AllowedTypes[strings.ToLower(t)] = struct{}{}
+		ca.fsWatcher.AllowedTypes[t] = struct{}{}
```
martinlindhe: This part was from earlier code I think, and looked correct. The filesystem types should be expressed in lower case (ext3, not Ext3 or EXT3)

fs.go
```diff
@@ -61,7 +61,7 @@ func (fw *FSWatcher) Results() (common.MeasurementsMap, error) {
 	}
 
 	for _, partition := range partitions {
-		if _, typeAllowed := fw.AllowedTypes[strings.ToLower(partition.Fstype)]; !typeAllowed {
+		if _, typeAllowed := fw.AllowedTypes[partition.Fstype]; !typeAllowed {
```
martinlindhe: This was already correct too I think

fs.go
troian: Oops. To much cut

PR #151, opened 2019-05-25
fix: verify sudo is present before using

troian:
run sudo in non-interactive mode
verify `service` command is available


pkg/hwinfo/hwinfo_nonwindows.go
```diff
@@ -17,14 +17,14 @@ import (
...
 
-func isCommandAvailable(name string) bool {
-	cmd := exec.Command("/bin/sh", "-c", "command", "-v", name)
-	if err := cmd.Run(); err != nil {
-		return false
-	}
-
-	return true
-}
+// func isCommandAvailable(name string) bool {
```
martinlindhe: Should be deleted

out.json
```diff
@@ -0,0 +1,2560 @@
+{
```
martinlindhe: I think this file was added in accident ?

troian: done

PR #153, opened 2019-05-25
fix: goreleaser `archive` is deprecated

troian:


troian:
Should be merged only after https://github.com/cloudradar-monitoring/docker-go-build/pull/1

martinlindhe:
strings.ReplaceAll was added in go 1.12 so https://github.com/cloudradar-monitoring/docker-go-build/blob/master/Dockerfile#L1 would need to be updated.

PR #155, opened 2019-05-26
use /etc/os-release if exists

troian:


pkg/osinfo/osinfo_linux.go
```diff
@@ -52,3 +134,24 @@ func osName() (string, error) {
 
 	return "", ErrUnknownOSType
 }
+
+func prettyId(id string) string {
```
golangcibot: func prettyId should be prettyID (from `golint`)

troian: done

PR #160, opened 2019-05-31
ignore cagent binary

martinlindhe:


.gitignore
```diff
@@ -1,4 +1,4 @@
 .idea
-cmd/cagent/cagent
+cagent
```
troian: windows as well?

martinlindhe: Done

PR #161, opened 2019-06-18
Logging adjustments

martinlindhe:
- We no longer write logs to stderr.
- Add `log_syslog = "syslog"` setting
- Disable ansi colors as they clobber log files
- @troian also reverted cagent gitignore that rules you brought up

DEV-1093


syslog.go
```diff
@@ -0,0 +1,41 @@
...
+	if syslogURL != "local" {
+		u, err := url.Parse(syslogURL)
+		if err != nil {
+			return fmt.Errorf("Wrong format of syslogURL: %s", err.Error())
+		}
+		network = u.Scheme
+		raddr = u.Host
+
+		if u.Port() == "" {
+			raddr += ":;514"
```
nikita-vanyasin: Please name the "514" const

martinlindhe:
>     1. There a lot of vendor files changed that are not relevant to the task solved by this PR.
>        Probably the `dep ensure -update` was issued causing the update of all dependencies.
>        I think this is kind of risky and should be moved to separate PR.

We did need updates to logrus dependency for this PR.

I will revert the vendor update and only update logrus.

>     2. Probably I've missed some discussions, but why the `log` alias for logrus package was removed? And why it was removed only from certain files, while all other parts of the project are still using `log`? Looks like `log` alias is common convention set by logrus maintainer. https://github.com/sirupsen/logrus

We decided to remove "log" alias because goimports auto resolve it to the internal "log" package, introducing hard to spot bugs. It has not been done in every file in this pr as it would add a ton of noise, but was brought over from frontman where it has been applied everywhere, since we copied syslog.go from there in this PR.

syslog.go
martinlindhe: done

martinlindhe:
@nikita-vanyasin I have reworked the vendor update to only include updated logrus

PR #168, opened 2019-07-01
Add the heartbeat: send GET request to HUB_URL every X seconds

nikita-vanyasin:
DEV-1116

New config variable is `heartbeat` (default 15).
Added validation for min interval values specified in config.

### DIFF
--- a/config.go
+++ b/config.go
@@ -17,6 +17,9 @@ import (
 const (
 	IOModeFile = "file"
 	IOModeHTTP = "http"
+
+	minIntervalValue          = 30.0
+	minHeartbeatIntervalValue = 5.0
 )
 
 var DefaultCfgPath string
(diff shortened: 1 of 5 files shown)


config.go
```diff
@@ -314,6 +319,14 @@ func (cfg *Config) validate() error {
 		}
 	}
 
+	if cfg.Interval < minIntervalValue {
+		return fmt.Errorf("interval value must be >= %.1f", minIntervalValue)
```
martinlindhe: For usability's sake, how about just force set the minimum allowed value and print a warning to the log?

config.go
```diff
@@ -314,6 +319,14 @@ func (cfg *Config) validate() error {
...
+	}
+
+	if cfg.HeartbeatInterval < minHeartbeatIntervalValue {
+		return fmt.Errorf("heartbeat value must be >= %.1f", minHeartbeatIntervalValue)
```
martinlindhe: Same

config.go
nikita-vanyasin: I think this isn't a good idea because most users don't have 'warning' log level enabled and they are not looking into log files in general. Accordingly, they will never know that there are such restrictions. Also, explicit is better than implicit. 
Maybe @thorstenkramm do have some thoughts on this, too?

martinlindhe: I didn't exactly mean log using the warning log-level, but just to add a line to the log about that we use default values because given value is out of range.

In multiple scenarios the user will not see the returned error anywhere and will be forced to consult log files with your proposed implementation just to figure out what went wrong.

For a end user I'm not sure I would agree hard failure on such trivial condition is good user experience, but I have no strong feelings either way.

PR #170, opened 2019-07-09
Add support for HWiNFO lib to retrieve temperature sensor values

nikita-vanyasin:
DEV-1023

PR #172, opened 2019-07-12
Updates in CI routine to support of proprietary MSI build

nikita-vanyasin:
Some more info here: DEV-1023

Changes include #170


nikita-vanyasin:
@martinlindhe I've also added some code related to proprietary build here https://bitbucket.org/cloudradar/win-build-deps/

PR #175, opened 2019-07-16
Exclude virtual network interface by default

nikita-vanyasin:
see DEV-1153

### DIFF
--- a/config_test.go
+++ b/config_test.go
@@ -3,6 +3,7 @@ package cagent
 import (
 	"io/ioutil"
 	"os"
+	"regexp"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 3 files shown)


config_test.go
```diff
@@ -179,3 +180,35 @@ fs_metrics = ['a', 'b']
...
+		excluded := false
+		for _, reString := range cfg.NetInterfaceExcludeRegex {
+			re, err := regexp.Compile(reString)
+			assert.NoError(t, err)
+			if re.Match([]byte(ifName)) {
+				excluded = true
+				break
+			}
+		}
+		assert.True(t, excluded, ifName + " excluded")
```
golangcibot: File is not `goimports`-ed (from `goimports`)
```suggestion
		assert.True(t, excluded, ifName+" excluded")
```

PR #178, opened 2019-07-19
Add support for three operation modes: full, minimal and heartbeat

nikita-vanyasin:
DEV-1148

### DIFF
--- a/handler.go
+++ b/handler.go
@@ -189,7 +189,7 @@ func (ca *Cagent) PostResultToHub(ctx context.Context, result *Result) error {
 	return nil
 }
 
-func (ca *Cagent) GetAllMeasurements() (common.MeasurementsMap, error) {
+func (ca *Cagent) CollectMeasurements(full bool) (common.MeasurementsMap, error) {
 	var errCollector = common.ErrorCollector{}
 	var measurements = make(common.MeasurementsMap)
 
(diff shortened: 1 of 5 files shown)


handler.go
```diff
@@ -188,99 +188,106 @@ func (ca *Cagent) PostResultToHub(ctx context.Context, result *Result) error {
...
-
-	ports, err := ca.PortsResult(processList)
-	errCollector.Add(err)
-	measurements = measurements.AddWithPrefix("listeningports.", ports)
-
-	swap, err := ca.SwapResults()
+	cpuUtilisationAnalysisResult, cpuUtilisationAnalysisIsActive, err := ca.CPUUtilisationAnalyser().Results()
 	errCollector.Add(err)
-	measurements = measurements.AddWithPrefix("swap.", swap)
+	measurements = measurements.AddWithPrefix("cpu_utilisation_analysis.", cpuUtilisationAnalysisResult)
```
golangcibot: `utilisation` is a misspelling of `utilization` (from `misspell`)

handler.go
```diff
@@ -188,99 +188,106 @@ func (ca *Cagent) PostResultToHub(ctx context.Context, result *Result) error {
...
+	if cpuUtilisationAnalysisIsActive {
+		measurements = measurements.AddWithPrefix(
+			"cpu_utilisation_analysis.",
```
golangcibot: `utilisation` is a misspelling of `utilization` (from `misspell`)

handler.go
nikita-vanyasin: This is all code that was sightly moved up. 
Can't fix spelling as it will break compatibility with backend.

handler.go
nikita-vanyasin: This is all code that was sightly moved up. 
Can't fix spelling as it will break compatibility with backend.

GolangCI checks are temporarely disabled.

nikita-vanyasin:
Rebased onto master

PR #179, opened 2019-07-25
Update 'bytes per second' calculation for windows.

nikita-vanyasin:
DEV-1039.

1) Update 'bytes per second' calculation for windows.
The old implementation could return incorrect value due to errors in underlying library (gopsutil).

2) Add config field to set fixed value for max network link speed.

### DIFF
--- a/config.go
+++ b/config.go
@@ -335,7 +335,7 @@ func (cfg *Config) GetParsedNetInterfaceMaxSpeed() (uint64, error) {
 		return 0, err
 	}
 	if value <= 0.0 {
-		return 0, fmt.Errorf("should be >= 0.0")
+		return 0, fmt.Errorf("should be > 0.0")
 	}
 
 	switch unit {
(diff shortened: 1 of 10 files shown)


config.go
```diff
@@ -318,6 +320,36 @@ func GenerateDefaultConfigFile(mvc *MinValuableConfig, configFilePath string) er
...
+		return 0, fmt.Errorf("can't parse")
+	}
+
+	valueStr, unit := v[0:len(v)-1], v[len(v)-1]
+	value, err := strconv.ParseFloat(valueStr, 0)
+	if err != nil {
+		return 0, err
+	}
+	if value <= 0.0 {
+		return 0, fmt.Errorf("should be >= 0.0")
```
martinlindhe: error message is inconsistent with check. maybe "should be > 0.0"

network.go
```diff
@@ -9,13 +9,15 @@ func (ca *Cagent) GetNetworkWatcher() *networking.NetWatcher {
 		return ca.netWatcher
 	}
 
+	maxSpeed, _ := ca.Config.GetParsedNetInterfaceMaxSpeed()
```
martinlindhe: maybe log error

PR #180, opened 2019-07-26
Improve operation_mode config comment

nikita-vanyasin:
To clarify that it applies only to io_mode=http


nikita-vanyasin:
Pretty trivial change, I'm merging the PR to fasten the process

PR #189, opened 2019-08-13
Update package build to recommend dependant tools instead of hard requirement

nikita-vanyasin:
Also:
- add dbus to the list of recommended packages: DEV-1177
- fix dmidecode detection. Now info message will be logged instead of an error

### DIFF
--- a/pkg/hwinfo/hwinfo_nonwindows.go
+++ b/pkg/hwinfo/hwinfo_nonwindows.go
@@ -17,8 +17,8 @@ import (
 	"github.com/cloudradar-monitoring/cagent/pkg/common"
 )
 
-func isCommandAvailable(name string) bool {
-	cmd := exec.Command("/bin/sh", "-c", "command", "-v", name)
+func isDmidecodeAvailable() bool {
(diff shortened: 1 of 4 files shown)


pkg/hwinfo/hwinfo_nonwindows.go
```diff
@@ -17,8 +17,8 @@ import (
 	"github.com/cloudradar-monitoring/cagent/pkg/common"
 )
 
-func isCommandAvailable(name string) bool {
-	cmd := exec.Command("/bin/sh", "-c", "command", "-v", name)
+func isDmidecodeAvailable() bool {
+	cmd := exec.Command("/bin/sh", "-c", "command -v dmidecode")
```
martinlindhe: Does this execute correctly? Arguments usually wants to be separated to exec.Command() in order for the shell to correctly parse the command invoked.

nikita-vanyasin: yep.

old code was wrong actually:
next argument to `sh -c` should be 'the command to execute'.

PR #200, opened 2019-09-09
Improve makefile syntax

nikita-vanyasin:


nikita-vanyasin:
It's actually a tiny change made with only purpose to create a trigger a new release.
There is a problem when building release with the same commit id as previous release

PR #203, opened 2019-09-15
Add storcli RAID monitoring support

nikita-vanyasin:
The new concept of 'modules' is introduced which allows to trigger alerts and warnings from the agent. The result is cached because we don't need to execute storcli as often.
The module is a standardized interface (see module.go)

More discussion on modules and feature here: DEV-1247


nikita-vanyasin:
@martinlindhe 
I've slightly updated the comments for config fields

PR #223, opened 2019-12-03
Jobmon tool

nikita-vanyasin:
The new binary allows running command and store it's output to special 'spool' directory.
Cagent will collect spool entries and send to Hub in a new field, removing the entry from spool on successful completion.

DEV-1324

A small refactoring was performed to have only one point of information about cagent package version.


pkg-scripts/postinstall-rpm.sh
```diff
@@ -7,6 +7,8 @@ CONFIG_PATH=/etc/cagent/cagent.conf
 versionsCount=$1
 
 if [ ${versionsCount} == 1 ]; then # fresh install
+    chown cagent:cagent -R /var/lib/cagent
+    chmod 6777 /var/lib/cagent/jobmon
```
requilence: what is the reason to have both `setgid` and `setuid`? jobmon can be run on a different user and push files with <666 permission?

example.config.toml
```diff
@@ -80,3 +80,12 @@ temperature_monitoring = true # default true
   #   binary = 'C:\Program Files\storcli\storcli64.exe'
   # Enable on Linux:
   binary = '/opt/storcli/sbin/storcli64'
+
+# Settings for the jobmon wrapper for the  monitoring of jobs
+[jobmon]
+  # Path to spool dir
+  #   spool_dir = 'C:\ProgramData\cagent\jobmon' # Windows
+  spool_dir = '/var/lib/cagent/jobmon' # Linux
```
requilence: what about darwin? `/var/lib` is a system-only directory. 
As I can see, you don't want tmp directory, so it should be something like `~/Library/Application Support/com.cloudradar.cagent/jobmon`

pkg-scripts/postinstall.sh
```diff
@@ -5,6 +5,8 @@ CONFIG_PATH=/etc/cagent/cagent.conf
 if [ "$1" = configure ]; then
     # $2 contains previous version number
     if [ -z "$2" ]; then # fresh install
+        chown cagent:cagent -R /var/lib/cagent
+        chmod 6777 /var/lib/cagent/jobmon
```
requilence: postinstall will run before the first run of jobmon, so `/var/lib/cagent/jobmon` may not exists? Should you `mkdir -p` it?

nikita-vanyasin: the folder will be created by package manager (DEB or RPM) before running post install script. The folders to be created are listed in goreleaser config file.

example.config.toml
nikita-vanyasin: Thanks, nice catch.

pkg-scripts/postinstall-rpm.sh
nikita-vanyasin: Yes, any user should be able to run jobmon, which will save the new file into that folder. Later cagent binary will read and remove it from folder. Also, there is one edge-case when jobmon can remove file created by other jobmon run, that exited unexpectedly.

pkg-scripts/postinstall.sh
requilence: what if the user uses the binary-only release(we still do it). Will `cagent` just fail because of the missing directory?

pkg/jobmon/spool.go
```diff
@@ -0,0 +1,256 @@
...
+type SpoolManager struct {
+	dirPath string
+	lock    *lockfile.Lockfile
+	logger  *logrus.Logger
+}
+
+// NewSpoolManager creates a new object to manage jobmon spool
+// dirPath must be absolute path
+func NewSpoolManager(dirPath string, logger *logrus.Logger) *SpoolManager {
+	lockFile, _ := lockfile.New(fmt.Sprintf("%s/spool.lock", dirPath))
```
requilence: in case the directory is missing, we will get nil dereference on line 204

pkg-scripts/postinstall.sh
nikita-vanyasin: nope, jobmon creates the directory if it does not exists.

Adding the directory to package-manager instructions allows to remove it automatically when removing package (with purge option).

pkg/jobmon/spool.go
nikita-vanyasin: lockfile is an external library.  `New()` call just constructs the object. Filesystem is not accessed

requilence: ah, I see. It just a wrapped string

pkg-scripts/postinstall.sh
requilence: got it 👍

example.config.toml
nikita-vanyasin: Fixed in a last commit

PR #224, opened 2019-12-17
Detect win proc states: suspended and not responding

requilence:
DEV-1338
- extract threads for each process to detect "suspended" processes
- query all windows with user32.dll to detect "not responding"(hanging) processes

### DIFF
--- a/pkg/winapi/ntdll.go
+++ b/pkg/winapi/ntdll.go
@@ -6,8 +6,10 @@ import (
 	"fmt"
 	"unsafe"
 
+	"github.com/davecgh/go-spew/spew"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
+	log "github.com/sirupsen/logrus"
 	"golang.org/x/sys/windows"
 )
 
@@ -25,6 +27,7 @@ var (
 )
 
 func checkNtDLLProceduresAvailable() error {
+	log.Debugf("checkNtDLLProceduresAvailable")
 	if ntDLLLoadErr != nil {
 		return errors.Wrap(ntDLLLoadErr, "winapi: can't load dll Ntdll.dll")
 	}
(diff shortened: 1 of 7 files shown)


pkg/winapi/ntdll.go
```diff
@@ -100,19 +102,19 @@ func GetSystemProcessInformation() (map[uint32]*SystemProcessInformation, error)
 	callWithBufferSize(currBufferSize)
 
 	if retCode != 0 {
-		logrus.Debugf(
-			"winapi call to NtQuerySystemInformation returned code: %d. required buffer size: %d. actual size: %d",
+		log.Debugf(
+			"[PROC] winapi call to NtQuerySystemInformation returned code: %d. required buffer size: %d. actual size: %d",
```
nikita-vanyasin: `[PROC]` log message part is not really related to pkg/winapi. (3 occurrences)


Also, logrus' recommended approach is to create a logger for each package and use it instead of direct calls to logrus. Such approach was already applied in some packages, I just haven't done it in winapi yet:
[pkg/monitoring/docker/docker.go#L13](https://github.com/cloudradar-monitoring/cagent/blob/e97b319a9a114323d2cb18ce81952bdb3f0c1ef3/pkg/monitoring/docker/docker.go#L13)
[pkg/monitoring/sensors/temperature.go#L14](https://github.com/cloudradar-monitoring/cagent/blob/6a35adfcaa39cb314660617c045ee77a7435b0c9/pkg/monitoring/sensors/temperature.go#L14)
[pkg/monitoring/raid/raid.go#L23](https://github.com/cloudradar-monitoring/cagent/blob/f9c7f26a7fc0f1f9ff51afd38f3e8c9309468dd9/pkg/monitoring/raid/raid.go#L23)

requilence: you are right. Fixed

PR #226, opened 2020-01-10
Allow to switch software RAID monitoring using config option 'software_raid_monitoring'

nikita-vanyasin:
DEV-1379

### DIFF
--- a/modules.go
+++ b/modules.go
@@ -22,7 +22,7 @@ func (ca *Cagent) initModules() {
 			return storcli.CreateModule(ca.Config.StorCLI.BinaryPath)
 		},
 		func() monitoring.Module {
-			return raid.CreateModule()
+			return raid.CreateModule(ca.Config.SoftwareRAIDMonitoring)
 		},
 	}
 
--- a/pkg/monitoring/raid/raid.go
+++ b/pkg/monitoring/raid/raid.go
@@ -24,11 +24,13 @@ var log = logrus.WithField("package", "raid")
 
 type RAID struct {
 	mdstatFilePath string
+	enabled        bool
 }
 
-func CreateModule() monitoring.Module {
+func CreateModule(enabled bool) monitoring.Module {
 	return &RAID{
 		mdstatFilePath: common.HostProc("mdstat"),
+		enabled:        enabled,
 	}
 }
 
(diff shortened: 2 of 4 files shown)


modules.go
```diff
@@ -22,7 +22,7 @@ func (ca *Cagent) initModules() {
 			return storcli.CreateModule(ca.Config.StorCLI.BinaryPath)
 		},
 		func() monitoring.Module {
-			return raid.CreateModule()
+			return raid.CreateModule(ca.Config.SoftwareRAIDMonitoring)
```
requilence: Why should it add the module in case it is not enabled?

pkg/monitoring/raid/raid.go
```diff
@@ -24,11 +24,13 @@ var log = logrus.WithField("package", "raid")
...
 type RAID struct {
 	mdstatFilePath string
+	enabled        bool
 }
 
-func CreateModule() monitoring.Module {
+func CreateModule(enabled bool) monitoring.Module {
 	return &RAID{
 		mdstatFilePath: common.HostProc("mdstat"),
+		enabled:        enabled,
```
requilence: I think it will be more clear not to have `enabled` bool inside the module. It makes no sense to have a disabled module added to the modules list

modules.go
nikita-vanyasin: The problem is different modules have different behavior of enabling/disabling them. I think it's good idea to encapsulate this behavior inside the module.
For example,  storcli module is enabled based on having non-empty "binary path" config value.

pkg/monitoring/raid/raid.go
nikita-vanyasin: Disabled modules are not added to list:
```
if m.IsEnabled() {
    modules = append(modules, m)
}
```
The first var contains only an enumeration of available modules and it's constructors.

Also see my previous comment regarding this situation.

modules.go
requilence: Ah, I see. Thanks for the explanation!

pkg/monitoring/raid/raid.go
requilence: Good, this way it seems ok

PR #228, opened 2020-01-13
Add monitoring of available updates

nikita-vanyasin:
apt-get, yum and dnf are supported

DEV-1333

### DIFF
--- a/pkg/monitoring/updates/apt.go
+++ b/pkg/monitoring/updates/apt.go
@@ -38,6 +38,8 @@ func (a *pkgMgrApt) FetchUpdates(timeout time.Duration) error {
 }
 
 func (a *pkgMgrApt) GetAvailableUpdatesCount() (int, error) {
+	// disable gosec G204 cmd audit:
+	/* #nosec */
 	cmd := exec.Command("sudo", a.GetBinaryPath(), "upgrade", "--dry-run")
 	out, err := cmd.Output()
 	if err != nil {
--- a/pkg/monitoring/updates/apt.go
+++ b/pkg/monitoring/updates/apt.go
@@ -1,0 +1,55 @@
+package updates
+
+import (
+	"context"
+	"fmt"
+	"os/exec"
+	"strings"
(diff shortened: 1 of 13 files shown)


pkg/monitoring/updates/apt.go
```diff
@@ -0,0 +1,55 @@
...
+		exitErr, ok := err.(*exec.ExitError)
+		if ok {
+			return errors.Wrap(exitErr, "while executing fetch command")
+		}
+	}
+	return err
+}
+
+func (a *pkgMgrApt) GetAvailableUpdatesCount() (int, error) {
+	cmd := exec.Command("sudo", a.GetBinaryPath(), "upgrade", "--dry-run")
```
golangcibot: G204: Subprocess launched with function call as argument or cmd arguments (from `gosec`)

requilence: any way we can avoid sudo? E.g. copy apt lists/package dir into /tmp, run `apt-get update` on it, then `apt-get upgrade --dry-run` or `dpkg --get-selections`

BTW, Not sure if this worth it

pkg/monitoring/updates/apt.go
```diff
@@ -0,0 +1,57 @@
...
+	"time"
+
+	"github.com/pkg/errors"
+)
+
+type pkgMgrApt struct {
+}
+
+func (a *pkgMgrApt) GetBinaryPath() string {
+	return "/usr/bin/apt-get"
```
requilence: does `apt-get` always available on new systems with just `apt`?

nikita-vanyasin: apt-get still ships with the all latest Debian-based systems.

Also, `man apt` on my Ubuntu 18 shows that
> apt provides a high-level commandline interface for the package management system. It is
       intended as an end user interface and enables some options better suited for interactive
       usage by default compared to more specialized APT tools like apt-get(8) and apt-cache(8).

PR #230, opened 2020-01-27
MySQL/MariaDB performance monitoring package

requilence:
DEV-1384

Metrics supported: 
- Updates per sec
- Inserts per sec
- Deletes per sec
- Queries per sec
- Commits pers sec
- Innodb data read bps
- Innodb data write bps
- Bytes Read bps
- Bytes Write bps


pkg/monitoring/mysql/mysql.go
```diff
@@ -0,0 +1,160 @@
...
+	}
+
+	statusTime := time.Now()
+	status, err := getStatus(client)
+	if err != nil {
+		report.AddAlert(fmt.Sprintf("failed to get status: %s", err.Error()))
+		return []*monitoring.ModuleReport{&report}, nil
+	}
+
+
```
golangcibot: File is not `goimports`-ed with -local github.com/cloudradar-monitoring/cagent (from `goimports`)
```suggestion

```

pkg/monitoring/mysql/status.go
```diff
@@ -0,0 +1,128 @@
...
+			total.WriteBytes = valInt
+		}
+	}
+
+	return &total, nil
+}
+
+func fillResultsPerSecond(new *Status, old *Status, durationBetween time.Duration, result map[string]interface{}) {
+	sec := durationBetween.Seconds()
+	result["Selects per sec"] = roundUp(onlyPositive(float64(new.Selects-old.Selects) / sec))
```
nikita-vanyasin: roundIUp and onlyPositive functions are used only in pair, therefore can be merged into one.

pkg/monitoring/mysql/status.go
```diff
@@ -0,0 +1,128 @@
...
+}
+
+func onlyPositive(f float64) float64 {
+	if f < 0 {
+		return 0
+	}
+	return f
+}
+
+func roundUp(p float64) float64 {
```
nikita-vanyasin: You probably can use common.RoundToTwoDecimalPlaces function to achieve the same result

pkg/monitoring/mysql/status.go
```diff
@@ -0,0 +1,128 @@
...
+		"Innodb data read bps":  nil,
+		"Innodb data write bps": nil,
+		"Bytes read bps":        nil,
+		"Bytes write bps":       nil,
+		"Queries per sec":       nil,
+	}
+}
+
+func onlyPositive(f float64) float64 {
+	if f < 0 {
```
nikita-vanyasin:
```suggestion
	return math.Max(f, 0)
```

pkg/monitoring/mysql/status.go
requilence: fixed

pkg/monitoring/mysql/status.go
requilence: Perfect. I've only found the unexported `roundUp` in other module

PR #231, opened 2020-01-29
Add configuration options for process monitoring to allow filtering and limits

nikita-vanyasin:
DEV-1387

Also made a small refactoring:  process monitoring related code moved to separate package.


nikita-vanyasin:
Rebased onto master tip. No new changes.

PR #235, opened 2020-02-11
Win updates reworked

requilence:
- Merge LinuxUpdatesChecks & WindowsUpdates into one package SystemUpdatesChecks
- Add the notice to the config that Mysql is experemental
- Add nil check to `MeasurementsMap.AddWithPrefix` and `MeasurementsMap.AddInnerWithPrefix`


pkg/monitoring/updates/watcher_windows.go
```diff
@@ -1,18 +1,22 @@
...
 	"time"
 
 	ole "github.com/go-ole/go-ole"
 	"github.com/go-ole/go-ole/oleutil"
-	log "github.com/sirupsen/logrus"
-
-	"github.com/cloudradar-monitoring/cagent/pkg/common"
 )
 
+type Config struct {
```
nikita-vanyasin: Why don't use the common updates.Config type both for windows and non-windows environments?
I know that one field is not needed for windows, but it's better than having a copy Config struct.

requilence: It is not used, forgot to remove

PR #259, opened 2020-06-15
Add "integration" test for cagent main function

nikita-vanyasin:
Test coverage increased from 16% to 35%

```
➜ go tool cover  -func=cover.out | grep total          
total:                                       (statements)                     16.1%
➜ go tool cover  -func=cover_new.out | grep total      
total:                                       (statements)                35.3%

```

### DIFF
--- a/handler_test.go
+++ b/handler_test.go
@@ -1,0 +1,39 @@
+package cagent
+
+import (
+	"io/ioutil"
+	"os"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func helperCreateCagent(t *testing.T) *Cagent {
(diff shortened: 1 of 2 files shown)


nikita-vanyasin:
> The test is failing in CI
> 


Yep, I've created PullRequest too early, sorry.

The reason new test failed was that `jaypipes/ghw` pkg was returning error when running in Docker image we use in CI routine.

That package actually has ugly design causing the trouble not for the first time. It's worth creating our own package for PCI devices detection without `ghw` flaws, but I think it's out of initial issue scope.

I've added if statement to ignore that error. That should also help end-users - they won't get false error message when running cagent.

PR #267, opened 2020-07-16
proper handling of submission interval

martinlindhe:
For all errors 400-599 - except 429 - wait until the next data submission is due according to the settings.
For 429 wait 10 seconds and try again

DEV-1601

### DIFF
--- a/handler.go
+++ b/handler.go
@@ -33,7 +33,7 @@ type cleanupCommand struct {
 }
 
 var (
-	ErrorHub429 = errors.New("Hub replied with a 429 error code")
+	ErrHubTooManyRequests = errors.New("Hub replied with a 429 error code")
 )
 
 func (c *cleanupCommand) AddStep(f func() error) {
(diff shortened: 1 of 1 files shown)


handler.go
```diff
@@ -32,6 +32,10 @@ type cleanupCommand struct {
 	steps []func() error
 }
 
+var (
+	ErrorHub429 = errors.New("Hub replied with a 429 error code")
```
nikita-vanyasin, line 36: Minor naming improvement: I think it's better to use a common well-known http status name instead of code, like ErrHubTooManyRequests

handler.go
```diff
@@ -278,6 +295,9 @@ func (ca *Cagent) sendHeartbeat() error {
 	}
 	req = req.WithContext(ctx)
 	resp, err := ca.hubClient.Do(req)
+	if resp != nil && resp.StatusCode == 429 {
```
nikita-vanyasin, line 298: 429 can be replaced with http.StatusTooManyRequests

handler.go
martinlindhe: Good idea. I'll do the same for corresponding changes in frontman shortly.

Fixed.

handler.go
martinlindhe: Fixed

PR #269, opened 2020-07-22
fix json encode error when sending to hub

martinlindhe:
This PR fixes the case where the "proc.list" Measurement contains Inf or NaN values. This happens if cagent is configured with `mem_monitoring = false`

JSON marshal cannot encode Inf or NaN values and errors out trying to send to hub.

example input that cause the issue:
```
&processes.ProcStat{PID:17355, ParentPID:995, ProcessGID:979, Name:"sleep", Cmdline:"sleep 60", State:"sleeping", Container:"", CPUAverageUsagePercent:0, RSS:0xc9000, VMS:0x71c000, MemoryUsagePercent:+Inf},
&processes.ProcStat{PID:15071, ParentPID:2, ProcessGID:0, Name:"kworker/0:7-events", Cmdline:"", State:"idle", Container:"", CPUAverageUsagePercent:0, RSS:0x0, VMS:0x0, MemoryUsagePercent:NaN},
            
```

DEV-1581 follow-up

### DIFF
--- a/pkg/monitoring/processes/processes_notwindows.go
+++ b/pkg/monitoring/processes/processes_notwindows.go
@@ -403,7 +403,10 @@ func gatherProcessResourceUsage(proc *process.Process, systemMemorySize uint64) 
 		log.WithError(err).Error("failed to get memory info")
 		return 0, 0, 0.0, 0.0
 	}
-	memUsagePercent := (float64(memoryInfo.RSS) / float64(systemMemorySize)) * 100
+	memUsagePercent := 0.0
+	if systemMemorySize > 0 {
+		memUsagePercent = (float64(memoryInfo.RSS) / float64(systemMemorySize)) * 100
+	}
 
 	// side effect: p.Percent() call update process internally
 	cpuUsagePercent, err := proc.Percent(time.Duration(0))


pkg/common/common.go
```diff
@@ -80,6 +80,9 @@ func MergeStringMaps(mapA, mapB map[string]interface{}) map[string]interface{} {
 }
 
 func RoundToTwoDecimalPlaces(v float64) float64 {
+	if math.IsNaN(v) || math.IsInf(v, 0) {
```
nikita-vanyasin, line 83: Maybe it's better to fix the underlying issue (incorrect values passed to the function) than masking it?
It's also unclear why we return 0 when value is `+Inf`. When speaking about MemoryUsagePercent it can be 100% and not 0, although I'm not sure how this happened.

martinlindhe: > Maybe it's better to fix the underlying issue (incorrect values passed to the function) than masking it?
Agree, lets figure out a better fix.

> It's also unclear why we return 0 when value is +Inf.
The input I saw, example above had both `MemoryUsagePercent:+Inf` and `MemoryUsagePercent:NaN`.

> When speaking about MemoryUsagePercent it can be 100% and not 0, although I'm not sure how this happened.

I explained the case in the first post. Let me add more detail.
If configured with `mem_monitoring = false`, then https://github.com/cloudradar-monitoring/cagent/blob/6f95682f7181bc284aa1deca7d72a265ad9caa12/pkg/monitoring/processes/processes.go#L68 will be called with memStat = nil, which leads to 

https://github.com/cloudradar-monitoring/cagent/blob/6f95682f7181bc284aa1deca7d72a265ad9caa12/pkg/monitoring/processes/processes.go#L77

with systemMemorySize = nil

martinlindhe: Dropped the old patch and sent a better fix.

PR #272, opened 2020-08-31
retries on http 5xx

martinlindhe:
Implements retry and give-up on hub Run() and RunHeartbeat() using a config like
```
on_http_5xx_retries = 4
on_http_5xx_retry_interval = 2.0
```

DEV-1668

### DIFF
--- a/config.go
+++ b/config.go
@@ -137,6 +137,9 @@ type Config struct {
 	FSMonitoring bool `toml:"fs_monitoring" comment:"Turn on/off any disk- and filesystem-related monitoring like fill levels and iops"`
 
 	NetMonitoring bool `toml:"net_monitoring" comment:"Turn on/off any network-related monitoring"`
+
+	OnHTTP5xxRetries       int     `toml:"on_http_5xx_retries" comment:"Number of retries if server replies with a 5xx code"`
+	OnHTTP5xxRetryInterval float64 `toml:"on_http_5xx_retry_interval" comment:"Interval between retries to contact server in case of a 5xx code"`
 }
 
 type ConfigDeprecated struct {
@@ -300,6 +303,9 @@ func NewConfig() *Config {
 		CPUMonitoring:    true,
 		FSMonitoring:     true,
 		NetMonitoring:    true,
+
+		OnHTTP5xxRetries:       4,
+		OnHTTP5xxRetryInterval: 2.0,
 	}
 
 	cfg.MinValuableConfig = *(defaultMinValuableConfig())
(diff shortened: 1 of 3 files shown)


config.go
```diff
@@ -137,6 +137,9 @@ type Config struct {
 	FSMonitoring bool `toml:"fs_monitoring" comment:"Turn on/off any disk- and filesystem-related monitoring like fill levels and iops"`
 
 	NetMonitoring bool `toml:"net_monitoring" comment:"Turn on/off any network-related monitoring"`
+
+	OnHTTP5xxRetries       int     `toml:"on_http_5xx_retries" comment:"Number of retries if server replies with a 5xx code"`
+	OnHTTP5xxRetryInterval float64 `toml:"on_http_5xx_retry_interval" comment:"Interval between retries to contact server in case of a 5xx code"`
```
nikita-vanyasin, line 142: I think having a measurement name in field comment would be great. It must be 'seconds' because all other fields uses it.

nikita-vanyasin (changes requested):
LGTM, but I'm not sure if having a fixed retry interval is really safe here.

in such use-cases it's common strategy to increase waiting time after each failed attempt. If it's not implemented, server can receive too much requests at one moment, resulting in hangs and even continuous failure. 
It's also common approach to increment interval progressively (e.g. +3 seconds, +10 seconds, +30 seconds, +2 min, etc.) and/or using random number generation (e.g. `+4 seconds and +math.Rand(0-10) seconds`)

Similar approach is already implemented in rport, maybe we can just reuse the code from there, although maybe it's overkill for cagent, better consult with Thorsten

martinlindhe:
> LGTM, but I'm not sure if having a fixed retry interval is really safe here.
> 
> in such use-cases it's common strategy to increase waiting time after each failed attempt. If it's not implemented, server can receive too much requests at one moment, resulting in hangs and even continuous failure.
> It's also common approach to increment interval progressively (e.g. +3 seconds, +10 seconds, +30 seconds, +2 min, etc.) and/or using random number generation (e.g. `+4 seconds and +math.Rand(0-10) seconds`)
> 
> Similar approach is already implemented in rport, maybe we can just reuse the code from there, although maybe it's overkill for cagent, better consult with Thorsten

martinlindhe:
> LGTM, but I'm not sure if having a fixed retry interval is really safe here.

Task is implemented as described in  DEV-1668. It does not contain any specs about a random retry interval.

config.go
martinlindhe: Fixed

nikita-vanyasin (approved):

PR #273, opened 2020-09-02
Adjust some error text and avoid some useless writes to log

martinlindhe:
Follow-up to DEV-1668

### DIFF
--- a/handler.go
+++ b/handler.go
@@ -71,7 +71,7 @@ func (ca *Cagent) Run(outputFile *os.File, interrupt chan struct{}) {
 				// for error codes 5xx, wait for configured amount of time and try again
 				retries++
 				if retries > ca.Config.OnHTTP5xxRetries {
-					log.Errorf("Run: hub connection error. giving up")
+					log.Error("Run: hub connection error. giving up")
 					return
 				}
 				log.Infof("Run: hub connection error %d/%d, retrying in %v s", retries, ca.Config.OnHTTP5xxRetries, ca.Config.OnHTTP5xxRetryInterval)
@@ -296,7 +296,7 @@ func (ca *Cagent) RunHeartbeat(interrupt chan struct{}) {
 				// for error codes 5xx, wait for configured amount of time and try again
 				retries++
 				if retries > ca.Config.OnHTTP5xxRetries {
-					log.Errorf("RunHeartbeat: hub connection error. giving up")
+					log.Error("RunHeartbeat: hub connection error. giving up")
 					return
 				}
 				log.Infof("RunHeartbeat: hub connection error %d/%d, retrying in %v s", retries, ca.Config.OnHTTP5xxRetries, ca.Config.OnHTTP5xxRetryInterval)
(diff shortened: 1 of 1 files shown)


handler.go
```diff
@@ -72,11 +71,13 @@ func (ca *Cagent) Run(outputFile *os.File, interrupt chan struct{}) {
 				// for error codes 5xx, wait for configured amount of time and try again
 				retries++
 				if retries > ca.Config.OnHTTP5xxRetries {
-					log.Errorf("hub connection error. giving up")
+					log.Errorf("Run: hub connection error. giving up")
```
m-terel, line 74: if no args please use `log.Error` instead.

handler.go
```diff
@@ -72,11 +71,13 @@ func (ca *Cagent) Run(outputFile *os.File, interrupt chan struct{}) {
...
 					return
 				}
-				log.Infof("hub connection error %d/%d, retrying in %v s", retries, ca.Config.OnHTTP5xxRetries, ca.Config.OnHTTP5xxRetryInterval)
+				log.Infof("Run: hub connection error %d/%d, retrying in %v s", retries, ca.Config.OnHTTP5xxRetries, ca.Config.OnHTTP5xxRetryInterval)
```
m-terel, line 77: Maybe would be better to use `Warn` instead of `Info`?

handler.go
```diff
@@ -287,21 +288,21 @@ func (ca *Cagent) RunHeartbeat(interrupt chan struct{}) {
...
 				time.Sleep(10 * time.Second)
 				continue
-			}
-			if err == ErrHubServerError {
+			} else if err == ErrHubServerError {
 				// for error codes 5xx, wait for configured amount of time and try again
 				retries++
 				if retries > ca.Config.OnHTTP5xxRetries {
-					log.Errorf("hub connection error. giving up")
+					log.Errorf("RunHeartbeat: hub connection error. giving up")
```
m-terel, line 299: if no args please use `log.Error` instead.

handler.go
```diff
@@ -287,21 +288,21 @@ func (ca *Cagent) RunHeartbeat(interrupt chan struct{}) {
...
 					return
 				}
-				log.Infof("hub connection error %d/%d, retrying in %v s", retries, ca.Config.OnHTTP5xxRetries, ca.Config.OnHTTP5xxRetryInterval)
+				log.Infof("RunHeartbeat: hub connection error %d/%d, retrying in %v s", retries, ca.Config.OnHTTP5xxRetries, ca.Config.OnHTTP5xxRetryInterval)
```
m-terel, line 302: The same, maybe `Warn`?

handler.go
martinlindhe: From issue description:
> Failed connections are logged with severity info. Only if all retries have failed, an error is logged.

handler.go
martinlindhe: Done

handler.go
martinlindhe: Done

handler.go
martinlindhe: See other comment

m-terel (approved):

PR #274, opened 2020-09-07
rework Run & RunHeartbeat retries to work correctly according to configured delays

martinlindhe:
DEV-1668 tweaks

### DIFF
--- a/handler.go
+++ b/handler.go
@@ -59,20 +59,23 @@ func (ca *Cagent) Run(outputFile *os.File, interrupt chan struct{}) {
 	}()
 
 	retries := 0
+	retryIn := secToDuration(ca.Config.Interval)
 
 	for {
 		err := ca.RunOnce(outputFile, ca.Config.OperationMode == OperationModeFull)
 		if err != nil {
 			if err == ErrHubTooManyRequests {
 				// for error code 429, wait 10 seconds and try again
-				time.Sleep(10 * time.Second)
-				continue
+				retryIn = 10 * time.Second
 			} else if err == ErrHubServerError {
 				// for error codes 5xx, wait for configured amount of time and try again
+				retryIn = time.Duration(ca.Config.OnHTTP5xxRetryInterval) * time.Second
 				retries++
 				if retries > ca.Config.OnHTTP5xxRetries {
-					log.Error("Run: hub connection error. giving up")
+					log.Error("Run: hub connection error")
 					retries = 0
+				} else {
+					log.Infof("Run: hub connection error %d/%d, retrying in %v s", retries, ca.Config.OnHTTP5xxRetries, ca.Config.OnHTTP5xxRetryInterval)
 				}
 			} else {
 				log.Error(err)
@@ -82,17 +85,8 @@ func (ca *Cagent) Run(outputFile *os.File, interrupt chan struct{}) {
 		select {
 		case <-interrupt:
 			return
-		default:
-			if retries > 0 {
-				log.Infof("Run: hub connection error %d/%d, retrying in %v s", retries, ca.Config.OnHTTP5xxRetries, ca.Config.OnHTTP5xxRetryInterval)
-				time.Sleep(time.Duration(ca.Config.OnHTTP5xxRetryInterval) * time.Second)
-				continue
-			}
-
-			select {
-			case <-time.After(secToDuration(ca.Config.Interval)):
-				continue
-			}
+		case <-time.After(retryIn):
+			continue
 		}
 	}
 }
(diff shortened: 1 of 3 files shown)


martinlindhe:
also updated github.com/shirou/gopsutil, golang.org/x/sys/unix dependencies hoping to resolve false AV match

handler.go
```diff
@@ -72,10 +72,8 @@ func (ca *Cagent) Run(outputFile *os.File, interrupt chan struct{}) {
 				retries++
 				if retries > ca.Config.OnHTTP5xxRetries {
 					log.Error("Run: hub connection error. giving up")
-					return
+					retries = 0
```
m-terel, line 75: Sorry, what is the reason to have `retries` if it's zerod here? And error tells that it's giving up, but it will continue to rerun, why is it done so?

Also, in line 69, is it expected that in case of `ErrHubTooManyRequests` it will rerun forever, even when a signal was received in `interrupt` channel?

handler.go
```diff
@@ -84,8 +82,17 @@ func (ca *Cagent) Run(outputFile *os.File, interrupt chan struct{}) {
 		select {
 		case <-interrupt:
 			return
-		case <-time.After(secToDuration(ca.Config.Interval)):
-			continue
+		default:
+			if retries > 0 {
+				log.Infof("Run: hub connection error %d/%d, retrying in %v s", retries, ca.Config.OnHTTP5xxRetries, ca.Config.OnHTTP5xxRetryInterval)
```
m-terel, line 87: This whole logic of retries seems to be complicated, and it seems it wouldn't work as expected. For example, if retries is already > 0 and an error that is != ErrHubTooManyRequests, != ErrHubServerError happened then it will wait for OnHTTP5xxRetryInterval interval, instead of ca.Config.Interval. But it's not expected behavior, if I understood it correctly.

m-terel (commented):
This PR contains 292 changed files...Please do not include vendor to it.

Please rewrite logic with retries, it seems it wouldn't work. Please cover with tests to be sure it works as expected.

handler.go
martinlindhe: > This PR contains 292 changed files...Please do not include vendor to it.

We version the vendor folder for this project. Please dont review the content of vendor folder.

martinlindhe: > Sorry, what is the reason to have retries if it's zerod here? And error tells that it's giving up, but it will continue to rerun, why is it done so?

The reasoning is laid out in issue DEV-1668.

```
Implement retries if cagent or csender gets 5xx from the hub.
Use a default config like

on_http_5xx_retries = 4
on_http_5xx_retry_interval = 2.0

Failed connections are logged with severity info. Only if all retries have failed, an error is logged.
```

handler.go
martinlindhe: Retries can only reach != 0 on ErrHubServerError. I don't understand how the corner case you describe could trigger.

handler.go
martinlindhe: > Please rewrite logic with retries, it seems it wouldn't work.

Please explain this further.

martinlindhe: > And error tells that it's giving up, but it will continue to rerun, why is it done so?

RunHeartbeat is a goroutine that runs for the duration of the server.

handler.go
m-terel: If I understood correctly:
1. run -> ErrHubServerError happend -> retries became 1 -> it slept for OnHTTP5xxRetryInterval
2. rerun -> err happend != ErrHubTooManyRequests and != ErrHubServerError. By current logic it will sleep for OnHTTP5xxRetryInterval and then rerun.

Shouldn't it sleep for ca.Config.Interval and then rerun in this case?

martinlindhe: We can extend the code further to handle the error cases like you lay them out, but it would only make the code more complicated and you just raised a comment about `this whole logic of retries seems to be complicated`.

In both cases delay will happen before a retry.

handler.go
martinlindhe: > This PR contains 292 changed files...Please do not include vendor to it.

vendor change was due to virustotal match discussed elsewhere. Since no success, it is reverted.

handler.go
martinlindhe: You're right. I'll rework the error check some more

m-terel: But behavior would be completely incorrect (both delay and output would be incorrect) in this scenario. It's a bug. It should be fixed. It can be rewritten in a new simple way, without making this complicated logic even more complicated. 

Please tell, what is the expected behavior of this code?

m-terel: @martinlindhe if you would like I can help you to fix it and make it more simple.

m-terel: Please consider to do something similar to this:
```
	retries := 0
	retryIn := secToDuration(ca.Config.Interval)

	for {
		err := ca.RunOnce(outputFile, ca.Config.OperationMode == OperationModeFull)
		if err != nil {
			if err == ErrHubTooManyRequests {
				// for error code 429, wait 10 seconds and try again
				retryIn = 10 * time.Second
			} else if err == ErrHubServerError {
				// for error codes 5xx, wait for configured amount of time and try again
				retryIn = time.Duration(ca.Config.OnHTTP5xxRetryInterval) * time.Second
				retries++
				// log error only when retries number exceeds a given threshold
				if retries > ca.Config.OnHTTP5xxRetries {
					log.Error("Run: hub connection error.")
					// set back to 0
					retries = 0
				} else {
					log.Infof("Run: hub connection error %d/%d, retrying in %v s", retries, ca.Config.OnHTTP5xxRetries, ca.Config.OnHTTP5xxRetryInterval)
				}
			} else {
				log.Error(err)
			}
		}

		select {
		case <-interrupt:
			return
		case <-time.After(retryIn):
			continue
		}
	}
```

martinlindhe: Updated PR with this suggestion

m-terel (approved):

PR #277, opened 2020-10-21
Log location for Unix is cagent.log

yamatt:
Previously the line stated that the logs go in to the configuration file on the Unix system. I suspect this is unlikely and a simple copy+paste error of the configuration section above.

This change updates the README to describe the true default destination of the logs.


martinlindhe:
Thanks for this!

PR #280, opened 2021-01-28
Improve jobmon spool multitasking

nikita-vanyasin:
DEV-1717

### DIFF
--- a/pkg/jobmon/spool.go
+++ b/pkg/jobmon/spool.go
@@ -219,13 +219,6 @@ func (s *SpoolManager) getLock(jobID string) error {
 		}
 	}
 
-	if err != nil {
-		if jobID != "" {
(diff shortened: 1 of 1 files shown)


pkg/jobmon/spool.go
```diff
@@ -95,14 +93,8 @@ func (s *SpoolManager) FinishJob(uniqID string, r *JobRun) error {
 }
 
 func (s *SpoolManager) GetFinishedJobs() ([]string, []*JobRun, error) {
-	err := s.getLock("")
```
martinlindhe, line 98: `s.getLock("")` was used as a lock for the jobs queue in order not to race between GetFinishedJobs() and RemoveJobs()

martinlindhe (approved):

